PR #13, opened 2015-10-19
Add IRI resolution tests

RubenVerborgh:
This pull request adds IRI resolution tests for Turtle and TriG, validating whether implementations correctly perform resolution as per RFC3986. [Discussion on the mailing list.](https://lists.w3.org/Archives/Public/public-rdf-comments/2015Aug/0000.html)


gkellogg:
+1 from me. [Call for consensus](https://lists.w3.org/Archives/Public/public-rdf-tests/2015Oct/0019.html) issued.

afs:
(Minor tweak) If changes are still being made, one test per `@base` per language would be nice.

afs:
The new test is marked  `rdft:TestTurtlePositiveSyntax` but should be `rdft:TestTurtleEval`

`rdft:TestTurtlePositiveSyntax` just requirtes the TTL file is parsable, no output checking.

RubenVerborgh:
@afs Thanks, corrected to `TestTurtleEval` and `TestTrigEval`.

Why would you prefer one test per `@base`? This would complexify the tests, it seems. What are the gain of doing so?

afs:
Same complexity - more entries in the manifest.

It makes finding a failing URI easier.

(Comparison of results should not depend on output order.)

RubenVerborgh:
I don't mind either way; I just don't want to attach too much importance on the IRI resolution tests. By splitting them, it might give the appearance that several topics are failing, while it's just one topic. Note that each entry has a unique URI as subject, which facilitates finding failing URIs.

Any other opinions on this? If several people agree, I happily split.

afs:
Issue about the application of RDF 3986 "sec 5.2.3 merge paths." and "sec 5.2.4: Remove Dot Segments" sent to the email list.  Not sure where this is best discussed.

https://lists.w3.org/Archives/Public/public-rdf-tests/2015Oct/0022.html

gkellogg:
Adding one test per base/relative IRI would create over 300 new tests where there are just about 300 tests right now, so about doubling the number of tests. My usual metric is about one test per feature/normative statement with duplications in some cases. I would be fine with having one test per base, rather than per base/relative combination; that's what we'd need for SPARQL anyway.

RubenVerborgh:
I see the case for alignment with SPARQL. Do I just number the tests then, i.e., `IRI-resolution-01` etc.?

gkellogg:
Yes, that seems reasonable. In general, naming tests using a standard naming convention, and not trying too much to stick to existing naming seems like a good way to separate work done by this group from the original work.

RubenVerborgh:
Alright, I split the tests. Made one exception though: I kept the miscellaneous tests (= not based on RFC3986) together; there are 3 `@base` statements in that file. Can change if required, but it seems to me that these tests belong together.

gkellogg:
@RubenVerborgh can we close this in favor of #30?

RubenVerborgh:
Yes, we might revisit cases 3–6 at a later time.

Issue #19, opened 2015-11-01
Update SPARQL 1.1 test suite to align with changes introduced in RDF 1.1


kasei:
RDF semantics surrounding "simple literals" changed the expected result of some tests in the SPARQL 1.1 Query test suite.

kasei:
This was resolved by the merge of #21.

PR #21, opened 2015-11-02
Removed function tests strdt03 and strlang03 to align with RDF 1.1 (#19)

kasei:
Removes tests :strdt03 and :strlang03 which fail under RDF 1.1 systems. Replaces them with :strdt03-rdf11 and :strlang03-rdf11 which use the same data and query, but updated expected result data based on RDF 1.1 semantics.

The SPARQL 1.1 test suite expects "abc"^^xsd:string and "DEF"^^xsd:string to produce an error when passed to STRDT and STRLANG, leading to ?str1 being unbound in these tests. This is because STRDT and STRLANG expect their first argument to a "simple literal". However, in RDF 1.1 "simple literals" are merely syntactic sugar for xsd:string datatyped literals. Therefore, both of these literals should be acceptable to STRDT and STRLANG, yielding values bound to ?str1.

### DIFF
--- a/sparql11/data-sparql11/functions/manifest.ttl
+++ b/sparql11/data-sparql11/functions/manifest.ttl
@@ -113,7 +113,6 @@
 	mf:name    "STRDT() TypeErrors (updated for RDF 1.1)" ;
 	mf:feature sparql:strdt ;
     dawgt:approval dawgt:NotClassified ;
-    dawgt:approvedBy <http://www.w3.org/2009/sparql/meeting/2012-01-31#resolution_3> ;
     mf:action
          [ qt:query  <strdt03.rq> ;
            qt:data   <data.ttl> ] ;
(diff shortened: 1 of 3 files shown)


gkellogg:
Are there other changes we should batch up? Pulling this requires sending out to public-rdf-tests@w3.org and public-sparql-dev@w3.org in a call for consensus, which I will perform. It would be good to optimize the attention of readers.

kasei:
I'm confident that these are the only changes relating to RDF 1.1 and SPARQL 1.1 Query. @afs may have thoughts on the other issues he mentioned, but I would think any changes related to those ought to be in a separate issue/PR anyway.

gkellogg:
[Call for consensus](https://lists.w3.org/Archives/Public/public-rdf-tests/2015Nov/0002.html) generated no negative feedback.

afs:
By the way, these changes are not to do with RDF 1.1 conformance.  They are just as wrong for SPARQL 1.1 / RDF 1.0.

PR #23, opened 2015-11-08
Corrections for tests plus-1 and plus-2

afs:
SPARQL 1.1 tests `functions/plus-1` and `functions/plus-2` include `ORDER BY`. Unfortunately, they sort strings against numbers.  This ordering is undefined by strict SPARQL. The results also don't identical query solutions adjacent in the ordering which is quite odd.

This PR corrects that by removing the `ORDER BY`.  The data and results are unchanged. `ORDER BY` is not a feature being tested.

Any system passing the original tests will still pass these corrected tests. 

The queries and the references in the manifest are renamed as `plus-1-corrected` and `plus-2-corrected`.

It may well be that systems are not checking the ordering when the query has `ORDER BY` in it.

Background: Apache Jena puts a total ordering on any `ORDER BY` with permitted extensions to compare literals with different datatypes (roughly, sort by lexical form then by datatype)

```
Adds queries plus-1-corrected.rq, plus-2-corrected.
Removes plus-1.rq and plus-2.rq.
Results remain the same.
```

### DIFF
--- a/sparql11/data-sparql11/functions/manifest.ttl
+++ b/sparql11/data-sparql11/functions/manifest.ttl
@@ -32,8 +32,8 @@
 	:contains01
 	:starts01
 	:ends01
-	:plus-1
-	:plus-2
+	:plus-1-corrected
+	:plus-2-corrected
 	:md5-01
 	:md5-02
 	:sha1-01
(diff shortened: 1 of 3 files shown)


gkellogg:
LGTM, I'll circulate for concensus.

gkellogg:
I'm going to incorporate this into another PR, including these changes, and adding `dawgt:approval dagwt:Proposed`, which I'll add into an update to the vocabulary.

gkellogg:
Superseded by #24. Closed by https://lists.w3.org/Archives/Public/public-rdf-tests/2015Nov/0015.html.

Issue #20, opened 2015-11-02
Remove non-official tests from SPARQL 1.1 test suite


kasei:
Many tests included in the SPARQL 1.1 test suite are either not approved, or were approved and then removed by the WG from the manifest lists. These tests should be removed to avoid confusion for people new to the test suite.

gkellogg:
Note this caused `rif02` and `rif05` to be removed from the entailment tests, as they were not `dawgt:Approved`.

cc/ @afs

afs:
A visual scan of the removal commit looks OK.  It initially looked bad to remove tests labelled "approved" but the manifest list makes it clear they were later unapproved by the WG.

kasei:
@gkellogg – I've also been working on a PR for this issue. Wondering if you think it's important to keep all the non-approved tests commented out in the manifests instead of actually removing them and actually cleaning up the manifests?

https://github.com/kasei/rdf-tests/commits/feature-20

gkellogg:
@kasei I figured that keeping them commented out was safest, at least for now. In the future, seeing what tests had been there, but were removed may be important, which could be lost if we remove them.

IMO, let's keep them commented out, and consider a bigger cleanup later on.

kasei:
So not interested in my commits that remove (now-)unused files? e.g. https://github.com/kasei/rdf-tests/commit/ec7d6e55e3260410d05e6a16f238fd768f43c0e6

I'm of the opinion that the git history will capture this stuff just fine, and (if we resolve to keep the two separate) so will the old WG-approved test suite.

gkellogg:
Good point. @afs, you're opinion?

afs:
Neutral.

The ideal would be a separate place to document each change with rationale but that's unrealistic extra work.  Finding the output that gets people involved with less burden is also important.

But it comes down to what are we producing?

Case 1: Material for an implementer who wants to run "the tests". Maybe they used to run the original versions, may be not. A clean set is probably better. Also, this is where additional tests are good.

Case 2: The input to the next WG in an area (if it happens). Additional tests are a bit dubious here. They are opinions the WG might not share (the WG might have changed the spec on seeing the tests!).

For case 1, what matters most is the finished state of the suite. Commented tests can be confusing (work in progress? might they appear later?). For case 2, maximum history.

gkellogg:
[...]

PR #28, opened 2015-12-19
Remove non-official tests from SPARQL 1.1 test suite

kasei:
Removes manifest entries and unused files for non-official SPARQL tests (#20).


gkellogg:
Thanks Greg, looks great!

Issue #6, opened 2015-09-09
Add Turtle, TriG, RDF/XML and SPARQL tests to better validate relative IRI resolution


gkellogg:
This was suggested by @RubenVerborgh and discussed on [public-rdf-comments](https://lists.w3.org/Archives/Public/public-rdf-comments/2015Aug/0000.html). The [gist](https://gist.github.com/RubenVerborgh/39f0e8d63e33e435371a) has been updated to include evaluation tests for Turtle, JSON-LD, and RDF/XML, which can readily be translated for TriG and SPARQL.

RubenVerborgh:
Thanks @gkellogg, feel free to assign this issue to me.

gkellogg:
I don't have the permission to add you quite yet (which is a bit odd). Once you're added, you should be able to assign yourself. Thanks!

BTW, I think I own you a response relating to `urn:ex:s276`, which I haven't looked into.

gkellogg:
See PR #13 for Turtle and TriG tests.

RubenVerborgh:
About RDF/XML, I wasn't sure where to put those. The RDF/XML tests have a pretty specific structure.

gkellogg:
I've created [RDF/XML versions of these tests](https://github.com/ruby-rdf/rdf-rdfxml/blob/develop/spec/reader_spec.rb#L370-L742) for my own use, which can be added to the RDF/XML test suite (I also have [JSON-LD tests](https://github.com/ruby-rdf/json-ld/blob/develop/spec/reader_spec.rb#L129-L521)). If you don't get to it, @RubenVerborgh, I'll handle it later this week.

SPARQL's another matter.

RubenVerborgh:
Only thing that held me back on RDF/XML is the [weird folder structure](https://github.com/w3c/rdf-tests/tree/gh-pages/rdf-xml); wasn't sure where to put things. So I'd leave it up to you if that's okay.

gkellogg:
Also [RDFa tests](https://github.com/ruby-rdf/rdf-rdfa/blob/develop/spec/reader_spec.rb#L1784-L2155).

gkellogg:
I added RDF/XML tests to PR #13.

afs:
I was pointed at the following text in RFC 3987 5.3.2.4.  Path Segment Normalization
>   The complete path segments "." and ".." are intended only for use
>   within relative references
[...]

PR #30, opened 2016-01-05
Add IRI resolution tests (subset)

RubenVerborgh:
Same as #13, but only adds tests 1, 2, 7, and 8 as discussed in #6.


gkellogg:
The branch seems to be missing some Turtle files:

* IRI-resolution-03.ttl/nt
* IRI-resolution-04.ttl/nt
* IRI-resolution-05.ttl/nt
* IRI-resolution-06.ttl/nt

gkellogg:
(Looks like equivalent TriG files are also missing.

RubenVerborgh:
That's on purpose, but I forgot to edit the manifest. Will do!

RubenVerborgh:
Fixed. This pull requests exists so people can test the cases of #13 for which there is consensus, as [requested by @afs](https://lists.w3.org/Archives/Public/public-rdf-tests/2016Jan/0005.html).

gkellogg:
Those all work for me, +1 to merge after comments go through.

PR #31, opened 2016-01-13
Index update

### LINKED ISSUE #22: Rebuild index files in test directories
Each group has created its own mechanism for an HTML view of their tests. Rather than trying to apply different strategies for each different test, I propose using something based on what I did for the [CSVW test suite](http://w3c.github.io/csvw/tests/), which generates a JSON-LD version of the test manifest and uses this to build an index.html. A simple script could then rebuild all index files.

Alternatively, something using [Github Jekyll](https://github.com/jekyll/jekyll) would probably be pretty easy too.

gkellogg:
This fixes #22.

* Creates JSON-LD versions of manifests
* Uses Haml templates to build HTML+RDFa versions of manifests as index.html for each test directory.
* Updates SPARQL README to use as sparql index
* Adds comments to each protocol and graph protocol test entries in Markdown format to display in HTML version of manifest used as directory index.

JSON-LD and index files built using Rake task depending on Ruby RDF tools being installed.


gkellogg:
cc/ @kasei @afs @ericprud.

No real functional changes, but SPARQL bits could use a once-over. Try accessing using [htmlpreview](http://htmlpreview.github.io/?https://github.com/w3c/rdf-tests/blob/index-update/sparql11/README.html).

afs:
Having JSON-LD and TTL files for the manifests may confuse people as to why there are two.  If the JSON-LD is for the processing pipeline, maybe they should not put into the git repo, leaving the TTL files as the master copy.

kasei:
Looks good to me (except the unfortunate blank node renaming in dc1f8f0 which made it hard to evaluate changes).

I agree with @afs about the potential for confusion on turtle vs. json-ld, but don't do anything with json-ld, so don't feel I'm in a good position to weigh the potential upside.

gkellogg:
I can inline the JSON-LD manifests to be just part of the HTML pipeline without making them manifest. But, we've done this elsewhere, and many people seem to like working with a JSON representation of the manifest as it eases processing when used in a test runner. I'll update the processing pipeline and eliminate the transitional manifest.jsonld files.

I'm afraid there's not much I can do about Blank-node renaming. But, the only place I'm aware that this bites is when an mf:result is a BNode, rather than a container for a ut:data. It's rare, but semantically equivalent. The blank node renaming within the diff is simply because the JSON-LD serializer uses a deterministic algorithm for (re)generating BNode labels, which is affected by lexical ordering, which can vary depending on insertions or deletions.

gkellogg:
These changes eliminate the manifest.jsonld files, but there are some support files left, such as manifest-frame.jsonld and template.haml; these shouldn't cause confusion, though.

PR #33, opened 2016-01-22
Adds mf:result for three Turtle collection list tests

robstewart57:
Relevant tests:

turtle-syntax-lists-01
turtle-syntax-lists-02
turtle-syntax-lists-03

These were previously TestTurtlePositiveSyntax tests. This commit
changes these 3 tests to TestTurtleEval tests, and adds the
corresponding expected NTriples results:

turtle-syntax-lists-01.nt
turtle-syntax-lists-02.nt
turtle-syntax-lists-03.nt

### DIFF
--- a/turtle/manifest.ttl
+++ b/turtle/manifest.ttl
@@ -1577,25 +1577,28 @@
    mf:action    <turtle-syntax-struct-05.ttl> ;
    .
 
-<#turtle-syntax-lists-01> rdf:type rdft:TestTurtlePositiveSyntax ;
+<#turtle-syntax-lists-01> rdf:type rdft:TestTurtleEval ;
    mf:name    "turtle-syntax-lists-01" ;
    rdfs:comment "empty list" ;
    rdft:approval rdft:Approved ;
    mf:action    <turtle-syntax-lists-01.ttl> ;
+   mf:result    <turtle-syntax-lists-01.nt> ;
    .
 
-<#turtle-syntax-lists-02> rdf:type rdft:TestTurtlePositiveSyntax ;
+<#turtle-syntax-lists-02> rdf:type rdft:TestTurtleEval ;
    mf:name    "turtle-syntax-lists-02" ;
    rdfs:comment "mixed list" ;
    rdft:approval rdft:Approved ;
    mf:action    <turtle-syntax-lists-02.ttl> ;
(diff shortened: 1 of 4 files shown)


robstewart57:
This adds a mf:result for tests turtle-syntax-lists-01, turtle-syntax-lists-02 and turtle-syntax-lists-03, which ensures that parsers generates blank nodes appropriately. If this is suitable as a pull request, please do check that the NTriples `.nt` files turtle-syntax-lists-01.nt, turtle-syntax-lists-02.nt and turtle-syntax-lists-03.nt do correctly represent the expanded version of the short hand `( .. )` shorthand in the corresponding Turtle `.ttl` files.

gkellogg:
These all look good to me. I'd say that if it's useful for an implementation to have these as Evaluation tests, that's a good enough reason to add.

gkellogg:
We may want to consider renaming the tests and files, as they're no longer syntax tests, and this would be odd.

robstewart57:
@gkellogg thanks. I agree that the move from positive syntax tests to evaluation tests mean that the file naming conventions turtle-syntax-lists-* are not appropriate. I'm happy to resubmit a new PR combining https://github.com/w3c/rdf-tests/pull/33 and https://github.com/w3c/rdf-tests/pull/36 , with an evaluation test for turtle-syntax-lists-04, using the file names:

turtle-eval-lists-01.ttl
turtle-eval-lists-01.nt
turtle-eval-lists-02.ttl
turtle-eval-lists-02.nt
turtle-eval-lists-03.ttl
turtle-eval-lists-03.nt
turtle-eval-lists-04.ttl
turtle-eval-lists-04.nt
turtle-eval-lists-05.ttl
turtle-eval-lists-05.nt

I'll close PRs 33 and 36 before submitting a new one with these filenames.

robstewart57:
Also,

>  I'd say that if it's useful for an implementation to have these as Evaluation tests, that's a good enough reason to add.

Having these turtle list examples as evaluation, rather than syntax tests has been very useful to iron out bugs in my implementation of the specification. This is generally the case whenever a parser has to carry state, e.g. when the parser departs a blank node property list and a blank node subject can be popped from the parser's stack.  Evaluation tests check that not only the syntax parsing complies, but also the state is correctly carried and modified during parsing.

PR #57, opened 2019-04-10
Fix missing named data graph in SPARQL 1.1 constructwhere04

rubensworks:
The following query from test `constructwhere04` expects data from the graph with name `data.ttl`.

```sparql
PREFIX : <http://example.org/>

CONSTRUCT
FROM <data.ttl>
WHERE { ?s ?p ?o }
```

However, the manifest does not define any data for this test.

This PR adds the proper graph data for this tests, and places it inside a named graph.

### DIFF
--- a/sparql11/data-sparql11/construct/manifest.ttl
+++ b/sparql11/data-sparql11/construct/manifest.ttl
@@ -55,7 +55,8 @@
     dawgt:approval dawgt:Approved ;
     dawgt:approvedBy <http://www.w3.org/2009/sparql/meeting/2011-02-01#resolution_3> ;
     mf:action
-         [ qt:query  <constructwhere04.rq> ] ;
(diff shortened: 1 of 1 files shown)


afs:
IIRC the expectation on "FROM" was to load the file so the test manifest did not need to mention the file because it is a "FROM". No harm in doing so if that aligns with engine practice.

There seems to be a bit usage of "FROM" to mean "take a named graph from the target dataset".

ericprud:
Does that mean that this merge means we no longer test loading from `FROM`?

afs:
If nowhere else, it's in the 1.0 suite.

I think (all this was a long time ago) this test is really about syntax - something between the "CONSTRUCT" and "WHERE " for a CONSTRUCT-WHERE query. If so, it is covered by syntax-query/syntax-construct-where-02.rq

PR #59, opened 2019-05-20
Clarify comparison of literals in SPARQL test suite

### LINKED ISSUE #58: How to format decimals?
In regards to the formatting of decimals, how should spec-compliant engines handle trailing zero's in decimals? I find that the expected results in this test suite are inconsistent, i.e. sometimes they expect a trailing zero, sometimes they do not.

Some examples that have **no** trailing zero:
 - <www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#seconds>
 - <www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#ceil01>

Some examples that **do** have a trailing zero:
 - <www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#coalesce01>
 - Many of the aggregates <https://www.w3.org/2009/sparql/docs/tests/data-sparql11/aggregates/manifest.ttl>

Should this test suite be consistent in it's formatting? Either consistently have a trailing zero (in conformance with the [canonical representation](https://www.w3.org/TR/2004/REC-xmlschema-2-20041028/#decimal-canonical-representation), or consistently remove it.
Or should we not compare the outputs by string equality?

cc: @rubensworks

rubensworks:
Closes #58


afs:
Yes, https://www.w3.org/TR/rdf-concepts/#section-graph-equality defines same-term isomorphism; only blank nodes are remapped.  There is a level above basic RDF, called D-entailment, that considers "same value".

Tests should be marked if they are value-sensitive but usually is in the test as FILTERS -- `=` compared to `sameTerm`.

Test should be marked up if they are value sensitive but may be some got missed.

PS as changes get made, can we keep text to a length that's reviewable in github please?

gkellogg:
(In spite of saying I approved, my comment was that this should be closed without change).

ericprud:
I like the implementation advice in
> Note that testing whether two result sets are isomorphic is simpler than full graph isomorphism. Iterating over rows in one set, finding a match with the other set, removing this pair, then making sure all rows are accounted for, achieves the same effect.

As a pedant, I think that approach of testing graph isomorphism works for comparing result set with rows of (?s, ?p, ?o), i.e. graphs. I propose changing that first sentence to
> Result set isomorphism can be tested with this algorithm:

rubensworks:
The goal of this PR was to document the suggestion from @cygri and @afs:

> So, as Andy said, the test suite documentation should state that literals are compared by value. Canonicalising actual and expected value before comparing them in the test runner is one way to achieve comparison by value.

The consensus from the comments above seem to be that instead of comparing literals in _all_ tests like this, only tests that require this comparison should be marked with a flag.

I'm happy with either approach, so I'm fine with closing this PR (and creating a new one) if everyone agrees.

PR #62, opened 2020-05-02
Restore sparql11/data-r2/optional-filter to proper location.

gkellogg:
For some reason, the optional-filter directory was under data-r2/sort; this moves it back.


gkellogg:
Since data-r2/manifest.ttl and data-r2/manifest-evaluation.ttl both reference optional-filter, and not sort/optional-filter, I'm chalking this up to an error in the original integration. The [DataAccess/tests](https://www.w3.org/2001/sw/DataAccess/tests/data-r2/) page this originally came from lists optional-filter at this level.

afs:
Agreed.

Issue #66, opened 2020-10-19
SPARQL test suite missing coverage of XPath Constructor Functions


kasei:
It looks like the [original DAWG tests](https://github.com/w3c/rdf-tests/tree/gh-pages/sparql11/data-r2/cast) only covered testing that the values resulting from a cast had the correct datatype, not the expected value. So, for example, an implementation could say `xsd:boolean("0")` evaluated to `true` and still pass the test suite.

I'm working on a PR to add some tests to improve coverage in this area.

kasei:
@afs – This is partly motivated by finding that Jena seems to disallow casting numerics others than `xsd:integer` to `xsd:boolean`. I *think* that [F&0 2.0 §17.1.6](https://www.w3.org/TR/2007/REC-xpath-functions-20070123/#casting-boolean) is the normative reference here:

> * If ST is xs:float, xs:double, xs:decimal or xs:integer and SV is 0, +0, -0, 0.0, 0.0E0 or NaN, then TV is false.
> * If ST is xs:float, xs:double, xs:decimal or xs:integer and SV is not one of the above values, then TV is true.

If you agree, can you confirm that I've understood Jena's implementation of the boolean casting correctly?

afs:
I agree that Jena's implementation is not casting numerics correctly.

BTW - generally - can refer to F&O latest even if the spec if technically F&O 2.x because the desire in the SPARQL 1.2 issues is to update.

https://www.w3.org/TR/xpath-functions/#casting-boolean

kasei:
@afs – Yeah, I had been relying on 3.1, but wanted to link to 2.0 just in case there were any differences that might explain the behavior. They seem the same to me, but I haven't read them closely to spot any divergence.

kasei:
As #68 was merged, I'm going to close this.

PR #68, opened 2020-10-20
SPARQL tests for XPath constructor functions

kasei:
This adds SPARQL evaluation tests for casting values using the XPath constructor functions:

* `xsd:boolean`
* `xsd:integer`
* `xsd:decimal`
* `xsd:float`
* `xsd:double`
* `xsd:string`

It does not (currently) add tests for `xsd:dateTime`, or casting of derived datatypes, or any of the special values such as infinities, `NaN`, or signed zero.


kasei:
Created this as a Draft PR to solicit feedback of WIP regarding #66.

sparql11/data-sparql11/cast/cast-bool.rq
```diff
@@ -0,0 +1,11 @@
...
+PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
+SELECT
+	?a
+	?v
+	(xsd:boolean(?v) AS ?boolean)
+WHERE {
+    ?a :p ?v .
+}
+ORDER BY ?s
```
afs, line 11: Should that be "ORDER BY ?a"?

kasei: Oops. The ordering was only for my benefit while working through the tests, and should just be removed.

gkellogg:
After some tweaks, I reproduce the expected results except for some xsd:string casts; I'm not clear from looking at the XPath definition how the value should be represented.

For example: `:n04 :p 0.0 .` is expected to generate a solution with `?string="0"`, and it's not clear to me why it's not "0.0", presuming that `?v` is canonicalized for its string representation. If it is expected to be as stated, then other results fail (e.g., b03 `false` vs `0`).

The following are solutions I expect that differ (represented as s-expressions):

```
((solution ?a=http://example.org/n06 ?v=-1.0 ?string=-1.0)
 (solution ?a=http://example.org/n05 ?v=1.0 ?string=1.0)
 (solution ?a=http://example.org/n07 ?v=0E1 ?string=0.0E0)
 (solution ?a=http://example.org/n04 ?v=0.0 ?string=0.0)
 (solution ?a=http://example.org/n10 ?v=1E0 ?string=1.0E0)
 (solution ?a=http://example.org/n08 ?v=1E0 ?string=1.0E0)
 (solution ?a=http://example.org/n09 ?v=0E1 ?string=0.0E0))
```

kasei:
I'm open to the possibility that I've gotten some of these wrong.

For `:n04`, we're casting a decimal `0.0` to a string. F&O 3.1 (which is newer than the SPARQL 1.1 references) [§19.1.1](https://www.w3.org/TR/xpath-functions/#casting-to-string) says:

> If ST is xs:decimal, then: If SV is in the value space of xs:integer, that is, if there are no significant digits after the decimal point, then the value is converted from an xs:decimal to an xs:integer and the resulting xs:integer is converted to an xs:string using the rule above.

I read this to mean that `xsd:string(0.0)` should produce `"0"`.

gkellogg:
Thanks, that clarifies it. This is pertinent for some Notation-3 tests as well, where datatypes are cast as string using the `string:concatenation` builtin.

gkellogg:
Based on the XPath description, there are probably some more tweaks necessary:

* Decimal values are expressed as integers if there are no digits after the decimal point.
* Floats and Doubles as decimal if the absolute value is between `0.000001` and `1000000`. Positive or negative zero represented as `"0"` or `"-0"` respectively.

So, for example, n07 should probably be "0" instead of "0E0". and n11 "1.25" instead of "1.25E0", and we may want to probe some of the extremes a bit further.

kasei:
Yeah, I saw the 0.000001…1000000 text. I wasn't sure how far to push these tests, as I imagine many systems will fail as we go farther, and I'm not sure I want to sign up to implement tests for the entirety of the XSD datatypes!

gkellogg:
These are remaining test values that I think should be used for xsd:string:

```
((solution ?a=http://example.org/n07 ?v=0E1 ?string=0)
 (solution ?a=http://example.org/n08 ?v=1E0 ?string=1)
 (solution ?a=http://example.org/n09 ?v=0E1 ?string=0)
 (solution ?a=http://example.org/n10 ?v=1E0 ?string=1)
 (solution ?a=http://example.org/n12 ?v=-7.875 ?string=-7.875)
 (solution ?a=http://example.org/n11 ?v=1.25 ?string=1.25))
```

While the existing values are:

```
((solution ?a=http://example.org/n07 ?v=0E1 ?string=0E0)
 (solution ?a=http://example.org/n08 ?v=1E0 ?string=1E0)
 (solution ?a=http://example.org/n09 ?v=0E1 ?string=0E0)
 (solution ?a=http://example.org/n10 ?v=1E0 ?string=1E0)
 (solution ?a=http://example.org/n11 ?v=1.25 ?string=1.25E0)
 (solution ?a=http://example.org/n12 ?v=-7.875 ?string=-7.875E0))
```

gkellogg:
Okay, looks good now.

gkellogg (approved):

lisp:
as, without a siginficant motivation, these promotion rules introduce gratuitous asymmetries into the data model, i sought among the discussions contributing to the “xpath and xquery…” document for threads which might provide that motivation.
on the topic of conversion to string values i found only

     https://lists.w3.org/Archives/Public/public-qt-comments/2004Feb/0538.html

this expresses a concern as to the consequences of the conversion rules for user interface presentations.
in the context of query processor operations on data which represents stored rdf, this case does not convince.

kasei:
@lisp convincing or not, we agree that they are the documented semantics?

As we've done with some other tests, we could mark some of these tests with an appropriate `mf:requires` value.

lisp:
i agree that that behaviour agrees with a recommendation related to xml.
i do not think it should apply to an rdf processor - that is either to a store or to a sparql processor.

afs:
I think the tests are right. Areas of F&O I didn't know about so great to have these tests.

@lisps's link is interesting showing the use case. 

> to implement tests for the entirety of the XSD datatypes!

We're in a better place with systems to be passing the test suite and if that means being mindful of the implementation burden on every corner case than so be it, especially as XSD 1.1/2.0 diff on more noticeable things.

lisp:
the case indicates that the test might well apply to xml-rdf encoding of sparql responses, but should it apply to gsp responses - which do involve coercion to strings, or to internal operators?

kasei:
@lisp – This is only about explicit casting using the XPath constructor functions in a query. If you think there are other unaddressed issues in the serializations or GSP, I think that should be a separate issue.

lisp:
> This is only about explicit casting using the XPath constructor functions in a query. 

that is exactly where it concerns me.
why should it be appropriate for a type promotion intended to resolve a concern with externalization to affect the nature of data which is internal to the query process and/or to be written to the store, but in neither case to be externalized?

kasei:
> why should it be appropriate for a type promotion intended to resolve a concern with externalization to affect the nature of data which is internal to the query process and/or to be written to the store, but in neither case to be externalized?

I'm afraid I don't understand this question.

The type *change* (not necessarily a "promotion") is externalized if you go on to store the data in the dataset, or return it as a query result.

lisp:
i use the term _promotion_, by analogy to a compiler's treatment, because the rules govern change in one direction only, but _change_ is fine.

i use externalization to describe a process which transforms an internal representation to the external representation which is required by a media type. thus, the example of rdf-xml, which corresponds to the use case which motivated the change. in the context of processing within an rdf service, there will be other transformations, but the are _internal_.

sparql11/data-sparql11/manifest-all.ttl
```diff
@@ -11,6 +11,7 @@
 		<basic-update/manifest.ttl>
 		<bind/manifest.ttl>
 		<bindings/manifest.ttl>
+		<cast/manifest.ttl>
```
afs, line 14: There is a query specific manifest - could this added to `manifest-sparql-query.ttl` as well?

kasei: Yes, sure. I didn't even realize those files existed, since it's also broken down in this file.

rubensworks (approved):
After fixing a couple of issues, all tests pass on my end as well. All of the expected output makes sense to me.

gkellogg:
Other than @lisp, I think there's consensus to merge these tests. @lisp do you object? If so, what would need to change to make it work for you?

lisp:
my objection stands, as stated.
that should not prevent a community decision to approve them.
it might suggest that tests which concern externalization should be marked as such and/or should be refactored into tests for the respective media types rather than for internal operations within a sparql processor.

kasei:
@lisp I don't really understand your distinction here between internal and external. These functions are internal to the SPARQL processor. While the tests do involve the SPARQL XML serialization, they could just as well have been written to do all the test comparisons within the SPARQL query and involve the serialization only for indicating whether the test passed. So I don't think there's any necessary connection between these tests and media types/serializations.

lisp:
> These functions are internal to the SPARQL processor.

which is why they should not enforce behaviour which was intended to accomplish some effect for just xml serialization.

kasei:
> which is why they should not enforce behaviour which was intended to accomplish some effect for just xml serialization.

Are you suggesting that implementations should be able to use any lexical form they want in the cast, or that we should specify our own formatting rules that duplicate the work done in F&O? Or perhaps some third option?

lisp:
i would expect that they should use a lexical form which, in this case, is expected for decimal values.

gkellogg:
RDF settled on using XPath forms for canonical literals for XSD datatypes, as creating an alternative would potentially create a source is incompatibility. If SPARQL was to describe (potentially incompatibly now) on an alternate canonical form for the XSD-based literals, it would seem odd and require justification. Given that XML Schema Datatypes and XPath are fairly closely aligned, the use of XPath for settling the normal form seems obvious.

The choice of using XML Schema datatypes for the basic datatypes-literals in RDF goes back quite a long way, doubtless influenced by RDF/XML as the only real serialization form early on, and that's the history we've maintained ever sense.

I could imagine that some other datatype (say schama:double) could have potentially settled on a different normal form, but these functions are defined for XSD datatypes.

JSON Literal canonicalization chose to use [JCS](https://www.rfc-editor.org/rfc/rfc8785) for describing the canonical form of rdf:JSON literals, for which includes it's own normalization for numeric types, but it's pretty close to the same rules decided for xsd:double in XPath for literals which are simple numerics.

gkellogg:
I would say that this PR should be merged, as it is consistent with the spec, and @lisp has suggested it be merged, in spite of his objection, given consensus. Any change to that should probably come in sparql 1.2.

Without objection, I'll merge this PR in the next couple of days.

afs:
+1 to merging.

And +1 to @kasei's dedication in reading F&O!

PR #74, opened 2021-05-13
Replace the wrong type "typed-literal" by "literal" in JSON results

BorderCloud:
Replace the wrong type "typed-literal" by "literal" in JSON results (https://www.w3.org/TR/sparql11-results-json/)


gkellogg:
That seems reasonable to me. @afs?

afs:
LGTM

PR #75, opened 2021-05-14
xsd:string is the datatype by default and becomes implicit in SPARQL results srx srj

BorderCloud:
xsd:string seems to become  the datatype by default and so, it becomes implicit in sparql results.

Here, I propose to modify the SPARQL results of tests in order to be in coformity with the SPARQL results generate today by the classic SPARQL services.


gkellogg:
Closed in favor of #76.

PR #76, opened 2021-05-14
xsd:string is the datatype by default and becomes implicit in SPARQL results srx srj

BorderCloud:
xsd:string seems to become the datatype by default and so, it becomes implicit in sparql results.

Here, I propose to modify the SPARQL results of tests in order to be in coformity with the SPARQL results generate today by the classic SPARQL services.

### DIFF
--- a/sparql11/data-sparql11/json-res/jsonres01.srj
+++ b/sparql11/data-sparql11/json-res/jsonres01.srj
@@ -17,12 +17,12 @@
       {
         "s": { "type": "uri" , "value": "http://example.org/s3" } ,
         "p": { "type": "uri" , "value": "http://example.org/p2" } ,
-        "o": { "datatype": "http://www.w3.org/2001/XMLSchema#string" , "type": "literal" , "value": "bar" }
+        "o": { "type": "literal" , "value": "bar" }
       } ,
       {
         "s": { "type": "uri" , "value": "http://example.org/s4" } ,
         "p": { "type": "uri" , "value": "http://example.org/p4" } ,
-        "o": { "datatype": "http://www.w3.org/2001/XMLSchema#integer" , "type": "literal" , "value": "4" }
+        "o": { "type": "literal" , "value": "4" }
       } ,
       {
         "s": { "type": "uri" , "value": "http://example.org/s5" } ,
--- a/sparql11/data-sparql11/csv-tsv-res/csvtsv01.tsv
+++ b/sparql11/data-sparql11/csv-tsv-res/csvtsv01.tsv
@@ -1,7 +1,7 @@
 ?s	?p	?o
 <http://example.org/s1>	<http://example.org/p1>	<http://example.org/s2>
 <http://example.org/s2>	<http://example.org/p2>	"foo"
-<http://example.org/s3>	<http://example.org/p3>	"bar"^^<http://www.w3.org/2001/XMLSchema#string>
+<http://example.org/s3>	<http://example.org/p3>	"bar"
 <http://example.org/s4>	<http://example.org/p4>	4
 <http://example.org/s5>	<http://example.org/p5>	5.5
 <http://example.org/s6>	<http://example.org/p6>	_:b0
--- a/sparql11/data-sparql11/csv-tsv-res/csvtsv02.tsv
+++ b/sparql11/data-sparql11/csv-tsv-res/csvtsv02.tsv
@@ -1,7 +1,7 @@
 ?s	?p	?o	?p2	?o2
 <http://example.org/s1>	<http://example.org/p1>	<http://example.org/s2>	<http://example.org/p2>	"foo"
 <http://example.org/s2>	<http://example.org/p2>	"foo"		
-<http://example.org/s3>	<http://example.org/p3>	"bar"^^<http://www.w3.org/2001/XMLSchema#string>		
+<http://example.org/s3>	<http://example.org/p3>	"bar"
 <http://example.org/s4>	<http://example.org/p4>	4		
 <http://example.org/s5>	<http://example.org/p5>	5.5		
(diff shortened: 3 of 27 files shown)


gkellogg (approved):
Seems reasonable to me.

afs:
There is at least one major triple store that provides RDF 1.0, not RDF 1.1.

There are probably RDF1.0 data around on the web.

These tests would work in a SPARQL 1.1/RDF 1.0 as well system because the data files align with the result files. jsonres tests also involve ORDER BY so there may be other effects.

While I wouldn't go out of the way to retain RDF 1.0 compatibility for new work, I don't think making a change when there isn't a need (bug or necessary change) helps overall.

SPARQL client software still has to deal with explicit xsd:string to be compliant so these do check something.

Minor: the original EARL reports were with RDF 1.0.

gkellogg:
@BorderCloud based on @afs's response, I've marked this as "wont fix". If you agree, you can close the PR.

afs:
It does ask the question what is this for - this community could decide that this repo is only about "now", or should reflect the history.  There are good arguments for either.

gkellogg:
At some point, the test suite, if not the SPARQL specs, need to fully align with RDF 1.1, particularly with regards to plain literals and xsd:string handling.

the point of this repo is to curate these tests with regard to the current specs, but if a SPARQL CG would like to move the specs forwards, we could/should track the changes here. It wouldn’t make sense to fork the test suite again.

afs:
Agreed. The test suite here should be RDF 1.1 usable - not another fork! SPARQL 1.1 (and 1.0) is written with the anticipation that simple literals (SPARQL terminology) merge xsd:string. But it is good to test for the case of RDF 1.0 written data and not see tests as a showcase of preferred usage.

gkellogg:
The JSON-LD tests have a `specVersion` option to allow some tests to be made specific to either JSON-LD 1.0 or 1.1; without the option, they are valid for both versions. We could consider adding such manifest annotations. Tests which require a Plain Literal and xsd:string literal to be different cause a problem for those using RDF 1.1 semantics.

afs:
@gkellogg - I'm neutral.

The situation hasn't changed. There is at least one major triple store, OpenLink Virtuoso, that provides RDF 1.0. For them, the explicit datatype is necessary. I found at least one test that has some data with and some without explicit `xsd:string` and this is reflected in the test results.

In the light of RDF-star (RDF 1.2) WG starting, if the rdf-tests community wants to align to RDF 1.1 then make this change (which I have checked and it seems OK) and also #83.

To check where the community opinion is, I've raised an issue to collect votes #88 because I think we need to be completely clear as to the decision going forward.  Give this at least some business-days to conclude.

If the community wishes to maintain as RDF 1.1 (and soon RDF 1.2), merge this.

If the community wishes to retain the current position (roughly - on a case-by-case basis), we should not make changes that are unnecessary and close this PR unmerged.

TallTed:
> The test suite here should be RDF 1.1 usable - not another fork!

I am not sure what is meant by "not another fork".

I think having branches focused on RDF 1.0, 1.1, and 1.2 (and actually, a matrix of these with SPARQL 1.0, 1.1, and 1.2) would be quite useful, for reasons discussed in #88, #83 and related.

afs:
I've tagged the repository with `sparql-mixed-rdf-version-tests` so that if this is merged, there is a convenient "before" marker.

afs (approved):

sparql11/data-sparql11/json-res/jsonres01.srj
```diff
@@ -17,12 +17,12 @@
...
         "s": { "type": "uri" , "value": "http://example.org/s3" } ,
         "p": { "type": "uri" , "value": "http://example.org/p2" } ,
-        "o": { "datatype": "http://www.w3.org/2001/XMLSchema#string" , "type": "literal" , "value": "bar" }
+        "o": { "type": "literal" , "value": "bar" }
       } ,
       {
         "s": { "type": "uri" , "value": "http://example.org/s4" } ,
         "p": { "type": "uri" , "value": "http://example.org/p4" } ,
-        "o": { "datatype": "http://www.w3.org/2001/XMLSchema#integer" , "type": "literal" , "value": "4" }
+        "o": { "type": "literal" , "value": "4" }
```
Tpt, line 25: This introduces the implicit rule "if a literal value is a JSON number that looks like a `xsd:integer`, then it should be parsed as a `xsd:integer` literal". Is it intended? It is not specify anywhere at might knowledge.

If we want to introduce it, it might also be nice to add more tests to cover more cases (boolean, the numbers `0.1` (`xsd:decimal`?) `0.1e3` (`xsd:double`?), `0.` (??)...

afs: No, that's not intended. Thank you for catching that.

I missed it because `json-res` isn't run by `manifest-sparq11-query`.

`csv-tsv-res/csvtsv02.tsv` is damaged as well.

`manifest-all` contains remote tests, which need special setup and the entailment tests.

I wonder if it would be useful to have `manifest.ttl` that is the "all local" part -- query, update and the result set formats.

afs: The entailment tests includes one with complex OWL reasoning.

The idea of `manifest.ttl` is a single place for the majority of triplestores.

Or `manifest-results.ttl`.
Or both!

Tpt: Makes sense. +1 to `manifest.ttl` or something like `manifest-with-results.ttl`.

afs: I'll add it to the reorg.
I'm currently minded to have `manifest-results` and `manifest` includes `-query,`, `-update` and `-results`.

PR #77, opened 2022-03-04
Add bad syntax tests for bnode labels with extra colons

afs:


afs:
The productions for PN_CHARS_U are different between Turtle and N-Triples - this is visible in blank node labels.

N-triples is supposed to be subset of Turtle - it appears that it is N-triples that is wrong.

Turtle:
```
[164s] 	PN_CHARS_U 	::=   PN_CHARS_BASE | '_'
```
N-Triples:
```
[158s] PN_CHARS_U        ::=  PN_CHARS_BASE | '_' | ':'
```
This PR adds tests to `turtle/` and `ntriples/` assuming a correction for N-Triples.

gkellogg (approved):

afs:
I'll add copies to NQ and TriG.

It is looking increasing certain that NT and NQ are wrong and the intention was for there not to be `:` in the blank node label itself.

Let's wait to see if any parsers systems respond.

lisp:
> Let's wait to see if any parsers systems respond.

we accept it as n-quads, as that is what the syntax says.
we do not accept it as sparql, as that is what the syntax says.

Tpt:
[The "Rio" Rust implementation](https://github.com/oxigraph/rio) passes all these tests without change (we applied the Turtle version of the rule).

afs:
Any reason not to merge this?

cc @lisp

lisp:
> Any reason not to merge this?

in itself no.
i do admit, however, that we continue not to recognize this as a failure.
i note also, that the tool to which we refer clients as a neutral reference for syntax issues - rapper, is also not sensitive to the restriction.

gkellogg:
@afs, I'd say merge at your discretion.

afs:
See also https://www.w3.org/2001/sw/wiki/RDF1.1_Errata#erratum_30

PR #82, opened 2022-11-19
Removes reference to not existing tests

Tpt:
:dawg-graph-10 and :dawg-dataset-{09,10,12} are still included in the mf:entries lists but do not exist anymore

### DIFF
--- a/sparql11/data-r2/dataset/manifest.ttl
+++ b/sparql11/data-r2/dataset/manifest.ttl
@@ -18,10 +18,7 @@
 	:dawg-dataset-06
 	:dawg-dataset-07
 	:dawg-dataset-08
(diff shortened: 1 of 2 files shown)


afs:
PR pulled down and tested by running the manifest in those two directories.

Caveat : data-r2 is a historical copy of SPARQL 1.0 and not all tests in the overall suite work due to an unrelated issue is that SPARQL 1.1 aligned to XSD and RDF 1.1 Turtle changes to decimal requiring a digit after the decimal point.

Do we want to fix up in place or rename as data-sparql10 and fix to be legal Turtle/SPARQL 1.1. (I think the decimal is the only issue.)

afs (approved):

PR #86, opened 2023-01-01
Update iri01.srx

### LINKED ISSUE #85: Incorrect variable name order in resultset file for 'iri01' test
### Observed

According to [SPARQL Query Results XML Format (Second Edition)](https://www.w3.org/TR/rdf-sparql-XMLres/#head):

> The order of the variable names in the sequence is the order of the variable names given to the argument of the SELECT statement in the SPARQL query. If SELECT * is used, the order of the names is undefined.

In the [iri01 query](https://github.com/w3c/rdf-tests/blob/main/sparql11/data-sparql11/functions/iri01.rq) the order of the variable names is `?uri ?iri`.  But in the header of the [corresponding resultset file](https://github.com/w3c/rdf-tests/blob/main/sparql11/data-sparql11/functions/iri01.srx) the order is `?iri ?uri`.

### Expected

The resultset serialization file for the `iri01` test to use the correct variable name order, i.e. `?uri ?iri`.

afs:
This closes #85.

### DIFF
--- a/sparql11/data-sparql11/functions/iri01.srx
+++ b/sparql11/data-sparql11/functions/iri01.srx
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <sparql xmlns="http://www.w3.org/2005/sparql-results#">
 <head>
-	<variable name="iri"/>
 	<variable name="uri"/>
+	<variable name="iri"/>
 </head>
 <results>
 		<result>


gkellogg (approved):

gkellogg:
@afs, I'd say merge when you think appropriate.

afs:
> @afs, I'd say merge when you think appropriate.

OK.

PR #95, opened 2023-04-01
Add action to automatically update HTML versions of test manifests

gkellogg:
Causes an PR that updates a manifest.ttl (or similar) to update the corresponding HTML manifest and add a commit to the PR.


afs:
HTML files in test areas have lost the indenting.

afs:
While it would be good to have automation here, if I understand the setup, there are consequences.

* Acts on each commit added after the initial PR  e.g. responding to comments, leading to repeated commits of updates on the PR branch.
* This appears to act on "Draft"

What I have seen elsewhere, but haven't used, is triggering by label. Add a label "prepare for merge" that does the index.html generation (and removes the label).

gkellogg:
> HTML files in test areas have lost the indenting.

Yes, sadly, Haml discontinued the indenting; the only way to get it back would be to either use a different templating engine (e.g, ERB), or run through tidy. I've taken the view that the HTML is really designed for rendering, and the loss of indenting is not significant.

gkellogg:
> While it would be good to have automation here, if I understand the setup, there are consequences.
> 
> * Acts on each commit added after the initial PR  e.g. responding to comments, leading to repeated commits of updates on the PR branch.

Yes, of course if the result is no change, it doesn't end up committing anything, but it does go through and do the work each time.

> * This appears to act on "Draft"

There's probably a way to filter for drafts.

> What I have seen elsewhere, but haven't used, is triggering by label. Add a label "prepare for merge" that does the index.html generation (and removes the label).

The idea is, and the result should be, that any change to a manifest.ttl file (really /manifest*\.ttl/) will cause the resulting index.html files to be rebuilt. Mostly, intended for people making changes to the test manifests that don't don't need to worry about the presentation. Using something such as a "prepare to merge" label would require it to be added, and we'd likely miss many updates that should trigger the rebuild. Rebuilding that results in no change just wastes CPU cycles, but it runs fairly quickly and shouldn't impede the workflow.

An alternate to using a label would be to trigger on the absence of a label, and then add one to prevent future work. However, that could miss important changes. I think, perhaps, that the best change would be to prohibit running when marked as a draft PR and enable when no longer a draft, if that results in an event that would trigger an action. But, still there's a chance that builds would be lost.

The cost of running with every commit, is that it could make several different commits if a manifest file is affected by a subsequent commit, but that doesn't feel like a real problem to me. It may favor doing more squash merges.

gkellogg:
The main branch is used for GH Pages, so they are available for people using cloned repos.

An alternative used elsewhere is to only update files when committing to the main branch, which probably requires something different due to branch protection issues. The advantage of adding the commits to the PR is that it allows the author and reviewers to see the rendered result of changes without trusting that it will just be right when merged. Personally, I favor this.

I can probably update "paths" in the action description to only fire when changes are made to either template or manifest files, but this wouldn't really reduce the number of commits made, if manifests are changed in different comments in the PR branch.

I am continuing on updates to correct existing rendering issues and to improve the whitespace in the results.

gkellogg:
> > HTML files in test areas have lost the indenting.

The HTML should be better indented now. Also, there were a number of errors in resulting HTML, and missing files to properly generate the SPARQL versions.

sparql/sparql10/ask/index.html
```diff
@@ -57,46 +57,40 @@
...
+    <p><a href="http://www.w3.org/Consortium/Legal/ipr-notice#Copyright">Copyright</a> © 2010 <a href="http://www.w3.org/"><abbr title="World Wide Web Consortium">W3C</abbr></a><sup>®</sup> (<a href="http://www.csail.mit.edu/"><abbr title="Massachusetts Institute of Technology">MIT</abbr></a>, <a href="http://www.ercim.org/"><abbr title="European Research Consortium for Informatics and Mathematics">ERCIM</abbr></a>, <a href="http://www.keio.ac.jp/">Keio</a>), All Rights Reserved. W3C <a href="http://www.w3.org/Consortium/Legal/ipr-notice#Legal_Disclaimer">liability</a>, <a href="http://www.w3.org/Consortium/Legal/ipr-notice#W3C_Trademarks">trademark</a>, and <a href="http://www.w3.org/Consortium/Legal/copyright-documents">document use</a> rules apply.</p>
```
afs, line 60: This is no longer correct.

afs:
There is a lot of boilerplate before the content. Do we need such boilerplate on all the pages?

There is no SPARQL 12 (see #96), no likes to originals and the SPARQL 1.1 points to an unhelpful manifest (the sparql11 get sit right - the query/update/results manifests).

"Abstract" -> not a section?

"Contributing Tests" -> refer to a contributing page?
"The test manifests and entries are built automatically ..."  This isn't needed - the key point for contributing is the PR. 

"Distribution" -> a one line license statement with links would do.

"To contribute to a W3C Test Suite" -- overlaps with "Contributing Tests".

"Disclaimer" - if there are links to the license, is this really necessary? It is big, bold and in the center of the page!

The © dates are not current.

Copyright - there isn't copyright assignment in place so the copyright remains with the contributor.

sparql/sparql10/ask/index.html
gkellogg: Indeed, I'll change them to the correct terminology and links.

gkellogg:
IIRC, much of what is in the template.haml files, which generate the HTML, came from READMEs in the corresponding directories. This can, of course, be customized, as the process looks for the closest template.haml file to create the corresponding index.

The License on the repo uses the [LICENSES FOR W3C TEST SUITES](https://www.w3.org/Consortium/Legal/2008/04-testsuite-copyright.html). I think all the documentation makes clear that these tests are provided under that license.

As this PR is getting overloaded with many things, and was intended to kick off the build automation, maybe we should consider content issues separately.

afs:
The link is now: https://www.w3.org/Consortium/Legal/2023/w3c-testsuite-license.html
(found via [2008/04-testsuite-license](https://www.w3.org/Consortium/Legal/2008/04-testsuite-license.html))

Of the two licenses :-) we want the open source once.  [3-clause BSD License](https://www.w3.org/Consortium/Legal/2023/software-license) although because [BSD-3-Clause](https://spdx.org/licenses/BSD-3-Clause.html) names the copyright holder. I'm not sure what the CG agreement covers. And we're not formally publishing 9as in a CG report).

As for the README in the leaf directories, I'm not sure that we need the text in index.html.

A LICENSE in a directory is enough, with pointer to the file from HTML (nowadays, generally, it is usual that link-to-license is as good as copy of text.)

gkellogg:
> The link is now: https://www.w3.org/Consortium/Legal/2023/w3c-testsuite-license.html (found via [2008/04-testsuite-license](https://www.w3.org/Consortium/Legal/2008/04-testsuite-license.html))
> 
> Of the two licenses :-) we want the open source once. [3-clause BSD License](https://www.w3.org/Consortium/Legal/2023/software-license) although because [BSD-3-Clause](https://spdx.org/licenses/BSD-3-Clause.html) names the copyright holder. I'm not sure what the CG agreement covers.I don't think this is formally "publishing as a CG report".

The group was chartered to take curate test suites created in different Working Groups, my expectation would be that the licensing would be principally the same. As all of these test suites are going to be taken over by the RDF-star WG, one way or, another, we should probably figure out what the group policy would be. Maybe we can push this onto @pchamin?

> As for the README in the leaf directories, I'm not sure that we need the text in index.html.

Depending on how people come across the tests, they may not see the README files. Having both README and index.html is somewhat redundant, but they certainly shouldn't be in conflict.

> A LICENSE in a directory is enough, with pointer to the file from HTML (nowadays, generally, it is usual that link-to-license is as good as copy of text.)

Would you suggest that all the "### Distribution" and "### Disclaimer" sections be change to reference the repo LICENSE file?

> Once we have a plan™️ we could pass it by the W3C CG contact. It shouldn't be a problem but it is good to get a check of it for the long term.

+1, but again, I think we should get the GH action process in place before making many more such changes.

PR #99, opened 2023-05-01
SPARQL DELETE/INSERT: an update operation should not be able to read its own writes

Tpt:


lisp:
this appears to be directed towards the consequences of just a "delete insert operation".
if that is , in fact, the case, the subject could reflect that.

sparql/sparql11/delete-insert/delete-insert-alloween-problem.ru
afs, line 1: Filename is "alloween" but "halloween" in the manifest.

Tpt: Sorry for that. It's now fixed.

afs (approved):

PR #104, opened 2023-07-14
Reorg RDF tests

gkellogg:
This creates a top level "rdf" directory, and puts all RDF 1.1 tests under "rdf/rdf11". It also fixes build issues, simplifies and consolidates templates, and reduces boilerplate from the resulting output.

* Process rdfs:comment for markdown in templates.
* Don't sort manifest references in template.
* Change "RDF Working Group" to "RDF 1.1 Working Group".
* Add rdf/rdf11 top-level manifest and generate index.
* Update Top-level manifest for separate RDF directory structure.
* Add temporary symlinks for old RDF directory locations.

Note: the symlinks for old RDF directory locations will be removed in a future PR.

For w3c/rdf-star-wg#76.


rubensworks:
@RubenVerborgh FYI, the [manifest locations](https://github.com/rdfjs/N3.js/blob/main/package.json#L61-L68) for N3.js will have to be updated after this merge, otherwise the CI will start failing.
(Note to self: also necessary for https://github.com/rdfjs/rdfxml-streaming-parser.js)

rubensworks (approved):

afs (approved):

gkellogg:
I was thinking of adding symbolic links from the old directory location to the new, and advertise the change more widely before removing them. That should keep existing dependencies working, but just delays the time when things will break unless updated.

PR #105, opened 2023-07-17
Remove symlinks for previous test locations for RDF 1.1 test suites.

gkellogg:
The RDF 1.1 test suites were relocated, and the symbolic link links were left as a temporary measure to allow dependent organizations to update to the new locations. This PR targets removing those links.


afs (approved):
Mostly harmless.

PR #107, opened 2023-07-18
Update base IRI expectations 

### LINKED ISSUE #106: Test relative-IRI resolution base inconsistent with expected results
Some tests, e.g., `rdf-turtle/turtle-subm-01.ttl` expect to be evaluated against the original test-suite location `http://www.w3.org/2013/TurtleTests/`. The result files are consistent with the (previous) location in `https://w3c.github.io/rdf-tests/turtle/`, instead.

See https://github.com/w3c/rdf-tests/blob/4e2e64b5b4a859b7326018c7ae8fe0a0b4b7fcbb/rdf/rdf11/rdf-turtle/README#L34-L41.

Either results need to be changed back to be relative to `http://www.w3.org/2013/TurtleTests/`, or the new location of `https://w3c.github.io/tests/rdf/rdf11/rdf-turtle/`.

gkellogg:
… relative to https://w3c.github.io/rdf-tests/.

Fixes #106.

### DIFF
--- a/ns/test-manifest.ttl
+++ b/ns/test-manifest.ttl
@@ -52,6 +52,11 @@
     rdfs:range   rdf:List ;
     .
 
+:assumedTestBase rdf:type rdf:Property ;
+    rdfs:comment "The assumed base URI for tests" ;
+    rdfs:domain	 :Manifest ;
+    .
+
 ## ---- Property declarations for each test ----
 
 :name rdf:type rdf:Property ;
--- a/rdf/rdf11/rdf-mt/README
+++ b/rdf/rdf11/rdf-mt/README
@@ -36,8 +36,8 @@ and the result is correct in the configured entailment regime with the
 configured recognized datatypes.
  
 
-The home of the test suite should eventually be
-<http://www.w3.org/2013/rdf-mt-tests/>. 
+The home of the test suite is
+<https://w3c.github.io/rdf-tests/rdf/rdf11/rdf-mt/>. 
 Per RFC 3986 section 5.1.3, the base IRI for parsing each file is the
 retrieval IRI for that file, but changing base IRIs should not affect any
 testing results.
--- a/rdf/rdf11/rdf-n-quads/README
+++ b/rdf/rdf11/rdf-n-quads/README
@@ -17,7 +17,7 @@ tests have a name (mf:name) and an input (mf:action).
 • An implementation passes a negative syntax test if it fails to parse
   the input.
 
-The home of the test suite is <http://www.w3.org/2013/NQuadsTests/>.
+The home of the test suite is <https://w3c.github.io/rdf-tests/rdf/rdf11/rdf-n-quads/>.
 
 See http://www.w3.org/2011/rdf-wg/wiki/RDF_Test_Suites for more details.
 
--- a/rdf/rdf11/rdf-n-triples/README
+++ b/rdf/rdf11/rdf-n-triples/README
@@ -17,7 +17,7 @@ tests have a name (mf:name) and an input (mf:action).
 • An implementation passes a negative syntax test if it fails to parse
   the input.
 
-The home of the test suite is <http://www.w3.org/2013/N-TriplesTests/>.
+The home of the test suite is <https://w3c.github.io/rdf-tests/rdf/rdf11/rdf-n-triples/>.
 
 See http://www.w3.org/2011/rdf-wg/wiki/RDF_Test_Suites for more details.
 
--- a/rdf/rdf11/rdf-trig/README
+++ b/rdf/rdf11/rdf-trig/README
@@ -26,12 +26,12 @@ tests have an expected result (mf:result).
 • An implementation passes a negative syntax test if it fails to parse
   the input.
 
-The home of the test suite is <http://www.w3.org/2013/TriGTests/>.
+The home of the test suite is <https://w3c.github.io/rdf-tests/rdf/rdf11/rdf-trig/>.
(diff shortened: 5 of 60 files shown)


gkellogg:
It affected RDF/XML tests even more than Turtle and TriG.

cc/ @Tpt.

Tpt:
Thank you! I have tested this change with Oxigraph test runner and it worked perfectly!

gkellogg:
@Tpt even though I can't add you as a reviewer quite yet, I believe you can leave a review by going to the "Files changed" tab.

Tpt (approved):
Indeed. Thank you!

afs:
Each test suite has an implicit base documented in the README but not available in the manifest. This makes writing a general purpose (i.e. independent of the test manifest being run) test runner impossible.

If the assumed base is in the manifest, then the tests can be downloaded and run. Running off the web in a development CI pipeline that is doing a lot of other build work is not always impractical.

This would be solved by putting the assumed home into the manifest so is machine accessible.:

```
   <> rdf:type mf:Manifest ;
      mf:assumedTestBase <https://w3c.github.io/rdf-tests/rdf/rdf11/rdf-turtle/> ;
      mf:entries (
        ...
```

afs (approved):

gkellogg:
I rebased this on the latest changes that use the `rdf/rdf11` directory structure.

I commented out the `assumedTestBase` entry in rdf-xml, as the existing tests assume that the location of the input file (`action`) is the base for resolving relative IRIs. If the tests are run against `http://w3c.github.io/rdf-tests/rdf/rdf11/rdf-xml/` it results in the appropriate relative IRI expansion. If run from a local directory the test runners need to act as if this is the URL used for retrieving the test actions.

Otherwise, we would either need to use a `assumedTestBase` for each test entry based on the `action` resource, or change all expected results to use the version for the manifest. Right now, I think the best thing is to not change the manifest.

My test runners correctly process Turtle, TriG, N-Triples, and N-Quads test manifests appropriately, and RDF/XML using the mechanism I described above.

afs (approved):
I suggest an issue for the RDF/XML case. Having the test runner hard coded with context is "quite inconvenient" when the same runner is used for other manifests.

gkellogg:
> I suggest an issue for the RDF/XML case. Having the test runner hard coded with context is "quite inconvenient" when the same runner is used for other manifests.

It really isn't hard coded with a context; as I mentioned above, it's based on the location of the input file. That's how my processor works, it just has a shim to proxy files based on the location in the test suite to whatever local directory I use.

gkellogg:
The situation would have been the same for the 1.1 test suite.  The RDF/XML tests have always been expected to run relative to the URL of the input file. We could, of course, do a comprehensive change to the RDF/XML tests that depend on this, but IIRC many RDF test suites take advantage of test locations for resolving relative IRI references.

afs:
>  IIRC many RDF test suites take advantage of test locations for resolving relative IRI references.

Outside RDF/XML, it's two tests in turtle and two tests in trig that depended on <https://w3c.github.io/rdf-tests/rdf/rdf11/>

There are other resolving tests but they are portable.

gkellogg:
I was thinking RDFa, N3, JSON-LD, and CSVW. Relative IRI/URL resolution is fairly common. What do you suggest we do to handle the RDF/XML tests? We could relocate the problematic tests to the base directory, for example.

afs:
The issue in RDF/XML is the subdirectories defeating mf:assumedTestBase. A quick grep - seems to be 20 NT files but 33 RDF files with rdf:ID in them.

Making assumed base work for reading the files might "just work" - i.e. setting the base externally as the file name. 

Any solution shouldn't make writing the manifest runner a rite-of-passage.

Do other test suites use the sub-directory pattern?

gkellogg:
> Making assumed base work for reading the files might "just work" - i.e. setting the base externally as the file name.

Prior to adding support for `mf:assumedTestBase` I used the location of the input file for resolving relative URLs. If run against the web location this works; if running against a local copy, it needs a shim/proxy to pretend it was loaded from the web location.

> Any solution shouldn't make writing the manifest runner a rite-of-passage.
> 
> Do other test suites use the sub-directory pattern?

Notation3 uses a [hierarchical test structure](https://github.com/w3c/N3/tree/master/tests/N3Tests) which includes at least one test for expanding relative IRIs [base-ref.n3](https://github.com/w3c/N3/blob/master/tests/N3Tests/cwm_syntax/base-ref.n3).

JSON-LD uses a [`base` property](https://github.com/w3c/json-ld-api/blob/a6850f2836623eb7b85260ca4de5c65f4a4e4474/tests/vocab.ttl#L172-L179) for each test entry that sets the base IRI to use.

RDFa used a [fairly flat structure](https://github.com/rdfa/rdfa.github.io/tree/master/test-suite/test-cases) but was dependent on a [web-based test runner](http://rdfa.info/test-suite/) which was quite difficult to maintain and made it challenging to run tests locally. (This is why I ended up writing my own SPARQL implementation).

CSVW had some [hierarchical tests](https://github.com/w3c/csvw/tree/gh-pages/tests), but I think tall those used for resolving relative IRIs were in the test directory, itself.

At this point, there are no RDF 1.2-specific tests needed for RDF/XML so no real issue for figuring this out for new tests. The old test hierarchy is pretty much unchanged from RDF 1.1, other than being in a new location. Not sure how much time we want to invest in this, unless it becomes a general issue for implementations. Certainly, running a test runner by pointing it directly at http://w3c.github.io/rdf-tests/rdf/rdf11/rdf-xml/ should _do the right thing_ at the expense of making running the tests from a local cache somewhat more challenging.

PR #109, opened 2023-07-28
Add assumedTestBase to give machine readable base URI information.

afs:
This is a PR to branch `fix-default-base`.

### DIFF
--- a/ns/test-manifest.ttl
+++ b/ns/test-manifest.ttl
@@ -52,6 +52,11 @@
     rdfs:range   rdf:List ;
     .
 
+:assumedTestBase rdf:type rdf:Property ;
+    rdfs:comment "The assumed base URI for tests" ;
+    rdfs:domain	 :Manifest ;
+    .
+
 ## ---- Property declarations for each test ----
 
 :name rdf:type rdf:Property ;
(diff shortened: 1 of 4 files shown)


Tpt (commented):
Thank you for pushing that. I believe we also need to set the base URI for `rdf-xml`.

afs:
Added for rdf-xml (but untested - the Jena build runs the WG test suite, not this copy.)

I intentionally didn't add it to `rdf-n-triples` and `rdf-n-quads` as it does not apply.
I didn't add it to `rdf-mt` because the README says it doesn't matter.

gkellogg (changes requested):
The [JSON-LD Expansion tests](https://github.com/w3c/json-ld-api/blob/main/tests/expand-manifest.jsonld) do this with a `baseIri` property on the manifest, but that's not really a well-defined term (it expands to `<https://w3c.github.io/json-ld-api/tests/vocab#baseIri>`.

I'm fine with `assumedTestBase`, but we should add an entry to https://github.com/w3c/rdf-tests/blob/main/ns/test-manifest.ttl describing the property to eventually by synchronized to http://www.w3.org/2001/sw/DataAccess/tests/test-manifest.

gkellogg (approved):
Added entry to the test manifest.

PR #117, opened 2023-10-23
N-Triples  base direction tests.

gkellogg:
Positive and Negative tests for the use of base direction in N-Triples.

### DIFF
--- a/rdf/rdf12/rdf-n-triples/syntax/ntriples-base-1.nt
+++ b/rdf/rdf12/rdf-n-triples/syntax/ntriples-base-1.nt
@@ -1,1 +1,1 @@
-<http://example/a>  <http://example/b> "Hello@en--ltr" .
+<http://example/a>  <http://example/b> "Hello"@en--ltr .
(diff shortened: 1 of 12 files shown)


rdf/rdf12/rdf-n-triples/syntax/ntriples-base-1.nt
```diff
@@ -0,0 +1 @@
+<http://example/a>  <http://example/b> "Hello@en--ltr" .
```
rubensworks, line 1: Should these looks as follows instead?
```suggestion
<http://example/a>  <http://example/b> "Hello"@en--ltr .
```

gkellogg: Yes, indeed.

rdf/rdf12/rdf-n-triples/syntax/ntriples-base-1.nt
```diff
@@ -0,0 +1 @@
+<http://example/a>  <http://example/b> "Hello"@en--ltr .
```
rubensworks, line 1: Minor nit: there are two spaces between `<http://example/a>` and `<http://example/b>`.
Not wrong, but a single space might be a bit nicer.

rubensworks (approved):
Looks good to me!

I'm not able to test it on my end though. I have to do some implementation work first.

rdf/rdf12/rdf-n-triples/syntax/ntriples-base-1.nt
gkellogg: Done.

Tpt (approved):

PR #119, opened 2023-11-01
Adds missing prefix definitions for bad syntax tests.

### LINKED ISSUE #9: missing prefixes in two negative syntax tests
http://www.w3.org/2013/TurtleTests/turtle-syntax-bad-missing-ns-dot-start.ttl
  http://www.w3.org/2013/TurtleTests/turtle-syntax-bad-missing-ns-dot-end.ttl

If the implementation doesn't fail on the intended error, it will fail on the missing prefix., c.f. http://www.w3.org/mid/20150719141712.GA9165@w3.org

gkellogg:
This has been around since 2015.

Fixes #9.

### DIFF
--- a/rdf/rdf11/rdf-trig/trig-syntax-bad-missing-ns-dot-end.trig
+++ b/rdf/rdf11/rdf-trig/trig-syntax-bad-missing-ns-dot-end.trig
@@ -1,1 +1,4 @@
+@prefix valid: <http://example.com/v#> .
+@prefix invalid.: <http://example.com/i#> .
+
 {valid:s valid:p invalid.:o .}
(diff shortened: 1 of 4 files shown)


domel (approved):

rdf/rdf11/rdf-trig/trig-syntax-bad-missing-ns-dot-end.trig
```diff
@@ -1 +1,4 @@
+@prefix valid: <http://example.com/v#> .
```
afs, line 1:
```suggestion
PREFIX valid: <http://example.com/v#>
```
etc etc.

Tpt (approved):

PR #122, opened 2023-11-30
GROUP BY test for SPARQL 1.2 to confirm SAMETERM semantics

hartig:
This test applies GROUP BY to literals to confirm that grouping is done based on SAMETERM semantics; as discussed in https://github.com/w3c/sparql-query/issues/131

### DIFF
--- a/sparql/sparql12/grouping/group-data-1.ttl
+++ b/sparql/sparql12/grouping/group-data-1.ttl
@@ -1,0 +1,7 @@
+@prefix : <http://example/> .
+@prefix xsd:    <http://www.w3.org/2001/XMLSchema#>
+
+:s1 :p "1"^^xsd:integer .
+:s2 :p "1"^^xsd:integer .
+:s3 :p "001"^^xsd:integer . 
+:s4 :p "1"^^xsd:string .
--- a/sparql/sparql12/grouping/group01.rq
+++ b/sparql/sparql12/grouping/group01.rq
@@ -1,0 +1,8 @@
+PREFIX : <http://example/>
+PREFIX xsd:    <http://www.w3.org/2001/XMLSchema#>
+
+SELECT ?v (COUNT(*) AS ?cnt)
+{
+  ?s :p ?v .
+}
+GROUP BY ?v
--- a/sparql/sparql12/grouping/group01.srx
+++ b/sparql/sparql12/grouping/group01.srx
@@ -1,0 +1,33 @@
+<?xml version="1.0"?>
+<sparql xmlns="http://www.w3.org/2005/sparql-results#">
+  <head>
+    <variable name="v"/>
+    <variable name="cnt"/>
+  </head>
+  <results>
+    <result>
+      <binding name="v">
+        <literal datatype="http://www.w3.org/2001/XMLSchema#integer">1</literal>
+      </binding>
+      <binding name="cnt">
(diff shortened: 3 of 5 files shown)


hartig:
@kasei you had a [comment about this test](https://github.com/w3c/sparql-query/issues/131#issuecomment-1819286212) which I am quoting here:

> New test looks good, but I think the manifest needs an `mf:requires` statement indicating that this test relies on no numeric canonicalization during loading. Some systems will fail this test simply because they canonicalize on import. (I thought we already had an IRI defined for such a `mf:requries` statement, but can't find anything in the existing manifest files.)

I am not sure what the object of this statement should be. Can you (or anyone else who knows how exactly to capture this) please create the relevant change suggestion in this PR.

sparql/sparql12/grouping/group-data-1.ttl
```diff
@@ -0,0 +1,7 @@
+@prefix : <http://example/> .
+@prefix xsd:    <http://www.w3.org/2001/XMLSchema#>
```
TallTed, line 2: Either...
```suggestion
@prefix : <http://example/> .
@prefix xsd:    <http://www.w3.org/2001/XMLSchema#> .
```

TallTed: Or...
```suggestion
PREFIX : <http://example/> 
PREFIX xsd:    <http://www.w3.org/2001/XMLSchema#>
```

kasei:
@hartig 

> I am not sure what the object of this statement should be. Can you (or anyone else who knows how exactly to capture this) please create the relevant change suggestion in this PR.

I don't think one exists yet. It would be something like:

```
:group01 mf:requires mf:NoXsdValueCanonicalization .
```

The naming could change depending on how wide/narrow we want such an annotation to be (right now this would be the only use of the annotation, but in the future, it might be useful to indicate no canonicalization of: numerics, xsd types, any known types, etc.).

hartig:
Makes sense @gkellogg. I will create such a manifest.ttl file once I am back at my computer.

hartig:
I just added `./sparql/sparql12/manifest.ttl` (see commit https://github.com/w3c/rdf-tests/pull/122/commits/9ff2e33a5e43e35a80e9ed3f48b5ddd4e4ec22e8)

hartig:
> > [@hartig] I am not sure what the object of this statement should be. Can you (or anyone else who knows how exactly to capture this) please create the relevant change suggestion in this PR.
>
> [@kasei ] I don't think one exists yet. It would be something like: [...]

Okay, thanks! Thinking that it would be better to not overload the discussion of this PR with a discussion of how the `test-manifest` vocabulary should be extended in order to capture such requirements, I have created a separate PR with the proposal for extending the vocabulary. See PR #123

rubensworks (approved):
Looks correct to me, I get the same output on my end.

hartig:
@gkellogg I have added `./sparql/sparql12/template.haml` (see commit https://github.com/hartig/rdf-tests/commit/57ac7492039001d013c82d575bef7863c0ac3165), which is a copy of `./sparql/sparql11/template.haml` with one of the Markdown lines changed to mention SPARQL 1.2 and the RDF-star WG.

If you see anything else that needs to be changed in this file, just make change suggestions in this PR.

gkellogg (approved):

PR #128, opened 2024-02-01
SPARQL 1.0: Covers equality of float and boolean

Tpt:
Both ware not covered


Tpt:
It seems that the CI does not support PRs from forks.

PR #129, opened 2024-02-01
Covers AVG behavior with empty groups

Tpt:
SPARQL 1.1: "If Count(S) = 0, then Avg(S) = "0"^^xsd:integer."


Tpt:
Replaced by #132

PR #131, opened 2024-02-01
SPARQL 1.1: Covers GROUP_CONCAT behavior with language tags

Tpt:
GROUP_CONCAT is defined on top of CONCAT and CONCAT has a well-defined behavior with language tag in RDF 1.1

### DIFF
--- a/sparql/sparql11/aggregates/agg-groupconcat-5.srx
+++ b/sparql/sparql11/aggregates/agg-groupconcat-5.srx
@@ -1,0 +1,7 @@
+PREFIX : <http://www.example.org/>
+ASK {
+	{SELECT (GROUP_CONCAT(?o) AS ?g) WHERE {
+		VALUES ?o { "1"@en "2"@fr }
+	}}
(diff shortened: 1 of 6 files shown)


sparql/sparql11/aggregates/agg-groupconcat-5.srx
```diff
@@ -0,0 +1,7 @@
+PREFIX : <http://www.example.org/>
```
rubensworks, line 1: Something probably went wrong here. Should be a results file.

Tpt: Sorry for that. It seems I got mixed up in my MRs and forgot to run this one with Oxigraph. Fixed and the tests pass with Oxigraph.

rubensworks (approved):
Thanks for the fix!

Test 4 doesn't pass on my end, but that's my mistake :-)

Tpt:
> Test 4 doesn't pass on my end, but that's my mistake :-)

Great news! The work to add and review these tests is not a waste!

PR #133, opened 2024-02-01
SPARQL 1.0: Covers equality of float and boolean

Tpt:
Both were not covered


rubensworks (approved):
I get the same results on my end.

PR #134, opened 2024-03-03
Updates N-Triples tests to use triple term syntax

gkellogg:
instead of quoted triples.

These tests use `rdf:reifies` and generally avoid triples which are not well-formed. Well-formedness tests likely needed later.


afs (approved):

TallTed:
Please change `N-Tripes` to `N-Triples` in PR title.

Also, please use `well-formed` instead of `wellformed` or `well formed`, wherever applicable.

domel (approved):

PR #148, opened 2024-10-22
GH-146: Remove resolution case with two possible answers

### LINKED ISSUE #146: Multiple correct outcomes for tests IRI-resolution-01 and IRI-resolution-02 (in Turtle and TriG)
[RFC 3986 section 5.2.2](https://datatracker.ietf.org/doc/html/rfc3986#section-5.2.2) allows for strict and non-strict resolution:

>      -- A non-strict parser may ignore a scheme in the reference
>      -- if it is identical to the base URI's scheme.

The tests
* rdf/rdf11/rdf-turtle/IRI-resolution-01.ttl
* rdf/rdf11/rdf-turtle/IRI-resolution-02.ttl

are affected because they have a test for resolving `http:g` against an http base. 

There are two correct answers: if `http:` is ignored, the result is the same as resolving `<g>` (non-strict) otherwise it is `<http:g>` (strict - scheme already present).

```
    if ((not strict) and (R.scheme == Base.scheme)) then
         undefine(R.scheme);
         endif;
    if defined(R.scheme) then
         T.scheme    = R.scheme;
         T.authority = R.authority;
         T.path      = remove_dot_segments(R.path);
         T.query     = R.query;
    . . .
```

IRI-resolution-02 has:
```
@base <http://a/bb/ccc/d/>.
...
<urn:ex:s044> <urn:ex:p> <g>.
...
<urn:ex:s084> <urn:ex:p> <http:g>.
```
IRI-resolution-01 has a base of `<http://a/bb/ccc/d;p?q>`.

afs:
* IRI-resolution-01 and IRI-resolution-02 in rdf-turtle and rdf-trig.
* Remove IRI-resolution work files (not in manifest).

This closes #146


gkellogg (approved):

kasei:
Change looks good, but why are IRI-resolution.nt and IRI-resolution.trig also removed as part of the PR?

kasei:
"Not in manifest," meaning entirely unused anywhere?

kasei (approved):

afs:
IRI-resolution.nt and IRI-resolution.trig were work files. They are all the cases in one file, with the base changing as originally created in the community. They got split into file with one file per base so each test is now "for this base ...".

PR #151, opened 2024-10-28
Add syntax tests for codepoint escaping.

kasei:
Adds new tests for some interesting cases of unicode codepoint escaping, addressing w3c/sparql-query#164. Two tests (codepoint-esc-01 and codepoint-esc-10) are marked in the manifest with `TODO` markers as being dependent on decisions on how systems should handle invalid escape sequences. I believe the others are accurately test the existing spec text of SPARQL 1.1.

I think many of these cases should also be turned into evaluation tests, to ensure the unescaping is being performed correctly, but I'll leave that for another PR (or a subsequent update to this PR).

### DIFF
--- a/sparql/sparql12/syntax-escaping/manifest.ttl
+++ b/sparql/sparql12/syntax-escaping/manifest.ttl
@@ -1,0 +1,81 @@
+@prefix rdf:    <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix :       <https://w3c.github.io/rdf-tests/sparql/sparql12/escaping#> .
+@prefix rdfs:   <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix mf:     <http://www.w3.org/2001/sw/DataAccess/tests/test-manifest#> .
+@prefix qt:     <http://www.w3.org/2001/sw/DataAccess/tests/test-query#> .
+@prefix dawgt:  <http://www.w3.org/2001/sw/DataAccess/tests/test-dawg#> .
+
+:manifest  rdf:type mf:Manifest ;
+    rdfs:label "SPARQL Codepoint-Escaping Tests" ;
+    mf:entries
+    ( 
+    :codepoint-esc-01
+    :codepoint-esc-02
+    :codepoint-esc-03
+    :codepoint-esc-04
+    :codepoint-esc-05
+    :codepoint-esc-06
+    :codepoint-esc-07
+    :codepoint-esc-08
+    :codepoint-esc-09
+    :codepoint-esc-10
+    ) .
+
+
+:codepoint-esc-01 rdf:type   mf:PositiveSyntaxTest ; # TODO: depends on decision on handling of invalid escapes
+   dawgt:approval dawgt:Proposed ;
+   mf:name        "codepoint-esc-01.rq" ;
+   rdfs:comment   "invalid codepoint escape (requires handling codepoint escaping before backslash escaping)" ;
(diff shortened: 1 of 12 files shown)


gkellogg:
If you create the branch for the PR in the rdf-tests repo, the automatic report generation should work properly. It's conceivable that there is a different package that allows pushing the changes to a remote repo, or some filter that would prevent running that action if the repo is not local.

gkellogg (approved):

afs:
Turtle handles Unicode escape sequences differently - it has `UCHAR` in the grammar and it can occur only in strings and URIs. Personally, I think this is a better design - a more common pattern, and it makes it clear what happens when the codepoint itself is meaningful near an escape sequences. I believe this should be "good practice".

The fact that obfusicated queries can be written in SPARQL is not good.
`\u0041\u0053\u004B\u0020\u007B\u007D` (`codepoint-esc-09.rq`)  (that's `ASK {}`)

And it is bad for streaming (SPARQL Update more than SPARQL Query).

The text [19.2 Codepoint Escape Sequences](https://www.w3.org/TR/sparql12-query/#codepointEscape) isn't precise how replacement happens. These are [errata](https://github.com/w3c/sparql-query/issues/164#issuecomment-2444195722) that need to be addressed in the spec..

We should split tests into two: "what we want", that is good practice (tobe agreed), and "full spec".

Surveying existing systems:
* Codemirror/YASGUI for SPARQL does not seem to support this.

kasei:
I've updated the PR so that it only contains tests that I think align with agreed upon spec text.

Despite what I think is agreement that we *shouldn't* support things like the obfuscated `\u0041\u0053\u004B\u0020\u007B\u007D` (renumbered in the latest update to :codepoint-esc-01), I think the spec is clear that this is currently supported and unambiguous.

I'll file a new issue with the two removed tests that probably need WG agreement before it is clear what the their results should be.

sparql/sparql12/syntax-escaping/manifest.ttl
```diff
@@ -0,0 +1,68 @@
+@prefix rdf:    <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix :       <https://w3c.github.io/rdf-tests/sparql/sparql12/escaping#> .
+@prefix rdfs:   <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix mf:     <http://www.w3.org/2001/sw/DataAccess/tests/test-manifest#> .
+@prefix qt:     <http://www.w3.org/2001/sw/DataAccess/tests/test-query#> .
+@prefix dawgt:  <http://www.w3.org/2001/sw/DataAccess/tests/test-dawg#> .
+
+:manifest  rdf:type mf:Manifest ;
+    rdfs:label "SPARQL Codepoint-Escaping Tests" ;
```
afs, line 9: We should indicate that these are related to the character stream processing. Calling them "escaping" is too general.

Suggestion:
1. Rename `syntax-escaping/` as `syntax-char-stream-processing/`
2. `rdfs:label "Character Stream Processing Tests" ;`

kasei: @afs I'll rename as requested, but unsure about the rdfs:label change. "Codepoint Escape" is a term used in the spec, but I don't think we use anything similar to "Character Stream Processing". I'm obviously biased as the author here, but I think without context I'd be confused by what "Character Stream Processing Tests" were, but would have a pretty good idea about "Codepoint-Escaping Tests". Thoughts?

Issue #46, opened 2017-06-08
SPARQL Tests for REGEX function


leipert:
Recently I had a look into the pattern definition of `REGEX` (and also `REPLACE`) to find out what "regex flavor" SPARQL [is using](https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#func-regex). The regex definition of [XQuery 1.0 and XPath 2.0 Functions](https://www.w3.org/TR/xpath-functions/#regex-syntax) and its base [XML Schema Part 2](https://www.w3.org/TR/xmlschema-2/#regexs) are quite interesting reads (and the latter is actually fun to read).

Unfortunately it seems like none of the SPARQL 1.1 libraries I tested (ARQ and RDFLib) use those definitions. They use the regex implementations of the underlying programming languages [Java.util.regex](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html) and [python's re](https://docs.python.org/3/library/re.html). Who would blame the developers? This leads to a mismatch between the SPARQL 1.1 specification and the implementation.

Apache Jena ARQ has 100% in the latest implementation report, while not supporting the `REGEX` function correctly. In my opinion tests for this function should be added in the next iteration of the test suite.

A few of the queries I tested (all taken from XQuery/XPath regex description):
-   Regex Flag `x` [(result should be yes, but errors)][1]:

    ```
    ASK {
      BIND("helloworld" AS ?s) .
      FILTER(REGEX(?s, "hello world", "x")) .
    }
    ```
-   Lowercase `\u212A` should match case insensitive `[A-Z]` [(result should be yes, but is no)][2]:

    ```
    ASK {
      BIND("\u212A" AS ?s) .
      FILTER(REGEX(?s,  "[A-Z]", "i")) .
    }
    ```
    
If I were to implement the tests, I would test, that...

- ... all flags (`s`, `m`, `i`, `x`) work correctly.
- ... faulty flags lead to errors
- ... [Character Classes](https://www.w3.org/TR/xmlschema-2/#charcter-classes) work as intended

[1]: http://www.sparql.org/books/sparql?query=++++ASK+%7B%0D%0A++++++BIND%28%22helloworld%22+AS+%3Fs%29+.%0D%0A++++++FILTER%28REGEX%28%3Fs%2C+%22hello+world%22%2C+%22x%22%29%29+.%0D%0A++++%7D&output=text&stylesheet=%2Fxml-to-html.xsl
[...]

PR #154, opened 2024-11-25
Add some tests covering regex quantifiers and options

Tpt:
Based on XPath spec: https://www.w3.org/TR/xpath-functions/#regex-syntax

Issue #46


Tpt:
Closing in favor of #155

PR #160, opened 2025-01-14
Improve coverage `GRAPH ?g {}`

Tpt:
`Graph <...> {}` should return a single empty solution if the graph exists and no solution if it does not


sparql/sparql10/graph/graph-empty-not-exist.rq
```diff
@@ -0,0 +1 @@
+SELECT * { GRAPH <unknown> {} }
```
rubensworks, line 1: `unknown` should probably become a valid IRI, such as
```suggestion
SELECT * { GRAPH <ex:unknown> {} }
```

afs: `<ex:unknown>`  is fine. Realistically, a URI scheme of `ex` meaning anything else is asking for trouble!

Aside --

* Registered URI scheme `example:` https://www.rfc-editor.org/rfc/rfc7595.html
* Registered URN namespace: `urn:example:` https://www.rfc-editor.org/rfc/rfc6963.html
* DNS TLD `.example` is defined and, unlike `example.org`, there is guaranteed to be no host.

Tpt: It is already a valid IRI because of the implicit base IRI. I have moved to `ex:unknown` to make it less confusing.

Tpt: Thank you for your reviews!

afs (approved):

rubensworks (approved):

PR #173, opened 2025-04-07
Add correct `xsd:string` types for some tests of SPARQL 1.1

joka921:
According to the SPARQL 1.1 standard (as specified by explicit examples there), the datatype `xsd:string` is to be preserved for string functions that have this datatype as an input, so a `SUBSTR`, `CONCAT` etc. of inputs with type `xsd:string` should yield results with `xsd:string`. The test suite currently assumes that the `xsd:string` datatype is dropped (strangely not only for the result of the expression, but also when simply exporting the typed input literals as XML.

This PR adjusts some tests (I have not yet fully analyzed all tests) to what I consider to be the more correct behavior. Please give me feedback especially if there is a particular reason why the datatype `xsd:string` is omitted in the expected query results (the SPARQL 1.1 standard explicitly differentiates between simple literals without a datatype, and literals explicitly typed with `xsd:string`, although they behave very similarly.

### DIFF
--- a/sparql/sparql11/functions/concat-single.srx
+++ b/sparql/sparql11/functions/concat-single.srx
@@ -5,7 +5,7 @@
 </head>
 <results>
 	<result>
-		<binding name="str"><literal>abc</literal></binding>
+		<binding name="str"><literal datatype="http://www.w3.org/2001/XMLSchema#string">abc</literal></binding>
 	</result>
 </results>
 </sparql>
--- a/sparql/sparql11/functions/concat01.srx
+++ b/sparql/sparql11/functions/concat01.srx
@@ -5,7 +5,7 @@
 </head>
 <results>
 		<result>
-			<binding name="str"><literal>abcDEF</literal></binding>
+			<binding name="str"><literal datatype="http://www.w3.org/2001/XMLSchema#string">abcDEF</literal></binding>
 		</result>
 </results>
 </sparql>
--- a/sparql/sparql11/functions/replace01.srx
+++ b/sparql/sparql11/functions/replace01.srx
@@ -23,11 +23,11 @@
 		</result>
 		<result>
 			<binding name="s"><uri>http://example.org/s5</uri></binding>
-			<binding name="new"><literal>abc</literal></binding>
+			<binding name="new"><literal datatype="http://www.w3.org/2001/XMLSchema#string">abc</literal></binding>
 		</result>
 		<result>
 			<binding name="s"><uri>http://example.org/s6</uri></binding>
-			<binding name="new"><literal>def</literal></binding>
+			<binding name="new"><literal datatype="http://www.w3.org/2001/XMLSchema#string">def</literal></binding>
 		</result>
 		<result>
 			<binding name="s"><uri>http://example.org/s7</uri></binding>
--- a/sparql/sparql11/functions/strafter01a.srx
+++ b/sparql/sparql11/functions/strafter01a.srx
@@ -27,7 +27,7 @@
 		</result>
 		<result>
 			<binding name="s"><uri>http://example.org/s6</uri></binding>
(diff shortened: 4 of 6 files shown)


afs:
> differentiates between simple literals without a datatype, and literals explicitly typed with xsd:string

A "simple literal" in RDF 1.1 is the way it is written. Simple literal `"abc"` does have a datatype of `xsd:string`. 

Writing the `datatype=` isn't necessary. Conventionally, it isn't included; e.g. it makes the results document bigger and in the migration from RDF 1.0  to RDF 1.1, the more common situation was use of RDF Plain literals (no datatype at all).

SPARQL 1.1 actually predates RDF 1.1 slightly so the spec has some explicit use of `xsd:string`.

The `rdf-tests` represent current community practice and have been updated for the implications of RDF 1.1. See commit #015c01e.

RDF 1.1 Concepts:

https://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal

> Please note that concrete syntaxes MAY support simple literals consisting of only a [lexical form](https://www.w3.org/TR/rdf11-concepts/#dfn-lexical-form) without any [datatype IRI](https://www.w3.org/TR/rdf11-concepts/#dfn-datatype-iri) or [language tag](https://www.w3.org/TR/rdf11-concepts/#dfn-language-tag). Simple literals are syntactic sugar for abstract syntax [literals](https://www.w3.org/TR/rdf11-concepts/#dfn-literal) with the [datatype IRI](https://www.w3.org/TR/rdf11-concepts/#dfn-datatype-iri) http://www.w3.org/2001/XMLSchema#string. Similarly, most concrete syntaxes represent [language-tagged strings](https://www.w3.org/TR/rdf11-concepts/#dfn-language-tagged-string) without the [datatype IRI](https://www.w3.org/TR/rdf11-concepts/#dfn-datatype-iri) because it always equals http://www.w3.org/1999/02/22-rdf-syntax-ns#langString.

joka921:
Hi @afs and thank you for the quick and detailed reply. I will try to quickly summarize how I understood it, so that you can confirm my understanding.
* In RDF 1.1, every literal is typed, concrete syntaxes might omit the specifications of `xsd:string` or (in case of a language tag) `langString`.
* The published version of SPARQL 1.1 specifies slight differences between "simple literals" and "typed literals with `xsd:string` which are incompatible to RDF 1.1, but can't be updated because of immutability of the specs document.
* In practice, the RDF 1.1 specification is used in the community also when dealing with SPARQL,  so `"abc"^^xsd:string"` and `"abc"` can and should be treated as equal by a SPARQL engine.

If you could quickly confirm my understanding, then this makes our life much easier. We are currently working on making our SPARQL engine [QLever](https://github.com/ad-freiburg/qlever) (more) conformant with SPARQL 1.1 and also automatically test against this test suite, so we stumbled upon this (currently our testing tool treats these mismatches as "yellow", so as a known but minor divergence, but with your update, we can set this mode to "green", so fully accepted.

afs:
Hi @joka921 - yes.

"abc"^^xsd:string" and "abc" are more than just equal. They are the same RDF term.

`sameTerm("abc", "abc"^^xsd:string)' is true and `{ ?x :p "abc" }` graph matches `:x :p "abc"^^xsd:string`.

TallTed:
> [`sameTerm("abc", "abc"^^xsd:string)' is true and `{ ?x :p "abc" }`graph matches`:x :p "abc"^^xsd:string`](https://github.com/w3c/rdf-tests/pull/173#issuecomment-2786097673).

Fixing markdown...

`sameTerm("abc", "abc"^^xsd:string)` is true and `{ ?x :p "abc" }` graph matches `:x :p "abc"^^xsd:string`.

afs:
@joka921 - do you agree that no change is necessary?

(cc @gkellogg)

PR #174, opened 2025-04-26
Tests for VERSION/@version in Turtle

afs:


gkellogg:
Looks good.

gkellogg (approved):

Tpt (approved):

rdf/rdf12/rdf-turtle/syntax/turtle-version-bad-04.ttl
```diff
@@ -0,0 +1 @@
+@version 1.2
```
ektrah, line 1: Should there be a `.` after the version specifier? (I.e., the example is bad because of the decimal token instead of the expected string token, not because of the missing `.` token.)

```suggestion
@version 1.2 .
```

rubensworks (approved):

rdf/rdf12/rdf-turtle/syntax/turtle-version-bad-04.ttl
afs: Good idea - seems reasonable to indicate the bad syntax is the `1.2` part.

PR #188, opened 2025-06-12
TriG test updates to match Turtle.

gkellogg:


Tpt:
I think you need to also update the `manifest.ttl` to point to the new files

Tpt:
Nothing is missing but the manifest is still pointing to the files you renamed under their old names.

gkellogg:
I seem to have not pushed up changes I had made locally, check again.

PR #190, opened 2025-06-16
Update RDF 1.2 semantics tests for triple terms not quoted triples.

gkellogg:
Updates tests which used the old quoted triples, typically in the subject position, to use triple terms in the object position.

### DIFF
--- a/rdf/rdf12/rdf-semantics/test000s.ttl
+++ b/rdf/rdf12/rdf-semantics/test000s.ttl
@@ -1,3 +1,3 @@
 prefix : <http://example.com/ns#>
 
-<< :a :b :c >> :p1 :o1.
+<<( :a :b :c )>> :p1 :o1.
--- a/rdf/rdf12/rdf-semantics/upercase-language-string.ttl
+++ b/rdf/rdf12/rdf-semantics/upercase-language-string.ttl
@@ -1,3 +1,3 @@
 prefix : <http://example.com/ns#>
 
-<< :a :b "hello"@en-US >> :p1 :o1.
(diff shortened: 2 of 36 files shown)


pfps (approved):
Looks great.  Merge it as the basis for the next step.

rdf/rdf12/rdf-semantics/test000s.ttl
```diff
@@ -1,3 +1,3 @@
 prefix : <http://example.com/ns#>
 
-<< :a :b :c >> :p1 :o1.
+<<( :a :b :c )>> :p1 :o1.
```
niklasl, line 3: It appears this should be removed (or moved to turtle and marked as negative/syntax error?).

rdf/rdf12/rdf-semantics/upercase-language-string.ttl
```diff
@@ -1,3 +1,3 @@
 prefix : <http://example.com/ns#>
```
niklasl, line 1: Nit-pick: s/upercase/uppercase/.

rdf/rdf12/rdf-semantics/test000s.ttl
gkellogg: This wasn't listed in entries, and there seem to be other which are defined and not listed. I'll remove this, and add the missing entries into `mf:entries`.

gkellogg: There were a number of tests which were not referenced in entries and the comment said that they were withdrawn as the current semantics didn't support the test. I removed the entries and test files.

Issue #193, opened 2025-06-23
Expand Syntax Tests of Decimal Numbers in Turtle


tobiaswjohn:
There are some tests that target different syntax for numbers in RDF 1.1 Turtle. However, not all valid syntax is covered. In particular, double / decimal numbers with a sign but no leading zero are missing. These cases could not be treated correctly by some tools in the past (see [Apache Jena Issue](https://github.com/apache/jena/issues/2715#ref-commit-a19c5a1)), so I think there addition would be valuable.
Example Turtle test files would be
```
<s> <p> -.2e3 .
```
and
```
<s> <p> +.7 .
```

afs:
@tobiaswjohn - Could you provide a PR?

Syntax tests go in:
https://github.com/w3c/rdf-tests/tree/main/rdf/rdf12/rdf-turtle/syntax

evaluation tests (which give the NT output) go in:
https://github.com/w3c/rdf-tests/tree/main/rdf/rdf12/rdf-turtle/eval

afs:
Either.

To some extent, rdf11 is the status at RDF 1.1 publication + some fixes.

An RDF 1.2 system must pass rdf11 and rdf12.

tobiaswjohn:
I will try to add the test cases and provide a pull request.

PR #194, opened 2025-06-24
Expand Syntax Tests of Decimal Numbers in Turtle

tobiaswjohn:
Addresses the issue raised in #193 by adding two test inputs for the syntax of decimal numbers in Turtle.

As suggested, I put the syntax tests into the RDF 1.2 folder, although the syntax are already part of RDF 1.1.

### DIFF
--- a/rdf/rdf12/rdf-turtle/syntax/index.html
+++ b/rdf/rdf12/rdf-turtle/syntax/index.html
@@ -358,6 +358,46 @@
             </dd>
           </dl>
         </dd>
+        <dt id='turtle12-number-1'>
+          <a class='testlink' href='#turtle12-number-1'>
+            turtle12-number-1:
+          </a>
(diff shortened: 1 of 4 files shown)


gkellogg (changes requested):
The tests themselves are fine, but as they're testing 1.1 behavior and not 1.2 behavior, I think they should be in the rdf11/rdf-turtle manifest. Generally, tests in rdf12/ are for new behavior or syntax expected with features introduced in RDF 1.2.

afs:
> The tests themselves are fine, but as they're testing 1.1 behavior and not 1.2 behavior, I think they should be in the rdf11/rdf-turtle manifest. Generally, tests in rdf12/ are for new behavior or syntax expected with features introduced in RDF 1.2.

Can we accept the PR then reorganise?

gkellogg (approved):
With a follow up to reorganize.

PR #200, opened 2025-06-27
fix some tests, mostly because triple terms are now transparent

pfps:
The tests were for an old semantics with opacity.

### DIFF
--- a/rdf/rdf12/rdf-semantics/manifest.ttl
+++ b/rdf/rdf12/rdf-semantics/manifest.ttl
@@ -82,7 +82,7 @@ trs:manifest a mf:Manifest;
   ) .
 
 trs:all-identical-triple-terms-are-the-same a mf:PositiveEntailmentTest;
-  rdfs:comment "Multiple occurences of the same triple terms are undistinguishable in the abstract model.";
+  rdfs:comment "Multiple occurences of the same triple term are the same domain element.";
   mf:action <test001a.ttl>;
   mf:entailmentRegime "simple";
   mf:name "all-identical-triple-terms-are-the-same";
@@ -92,7 +92,7 @@ trs:all-identical-triple-terms-are-the-same a mf:PositiveEntailmentTest;
   test:approval test:NotClassified .
 
 trs:annotated-asserted a mf:PositiveEntailmentTest;
-  rdfs:comment "Annotated triples are asserted.";
+  rdfs:comment "Annotated triples are asserted.  This is about shorthand expansion, and is not really a semantics test.";
   mf:action <test007a.ttl>;
   mf:entailmentRegime "simple";
   mf:name "annotated-asserted";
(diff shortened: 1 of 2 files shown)


gkellogg (approved):

rdf/rdf12/rdf-semantics/index.html
```diff
@@ -408,13 +408,13 @@ <h2>
           </a>
           <span about='../rdf-semantics#malformed-literal' property='mf:name'>malformed-literal</span>
         </dt>
-        <dd inlist='true' property='mf:entry' resource='../rdf-semantics#malformed-literal' typeof='mf:NegativeEntailmentTest'>
+        <dd inlist='true' property='mf:entry' resource='../rdf-semantics#malformed-literal' typeof='mf:PositiveEntailmentTest'>
```
doerthe, line 411: I am somehow lost here, sorry.  I cannot access the file https://w3c.github.io/rdf-tests/rdf/rdf12/rdf-semantics/malformed-literal.ttl and would like to see it to understand this test. Is this addressed in another PR?

rdf/rdf12/rdf-semantics/index.html
```diff
@@ -606,7 +606,7 @@ <h2>
         </dt>
         <dd inlist='true' property='mf:entry' resource='../rdf-semantics#opaque-iri-control' typeof='mf:PositiveEntailmentTest'>
           <div property='rdfs:comment'>
-            <p>Check that owl:sameAs works as expected, to ensure that opaque-iri does not pass spuriously.</p>
+            <p>Check that owl:sameAs works as expected, was to ensure that opaque-iri does not pass spuriously.</p>
           </div>
```
doerthe, line 610: Should we delete this test as owl:sameAs is not part of RDF? Or maybe list it somehow as a special case?

doerthe (commented):
I agree with the other changes.

rdf/rdf12/rdf-semantics/index.html
gkellogg: Note that this file is build automatically from rdf-semantics/manifest.ttl, so best make changes there. That will automatically rebuild the index.html.

TallTed: @pfps — Will you create a new PR against `rdf-semantics/manifest.ttl`, and carry the suggested revisions from myself and @doerthe there?

gkellogg: Might be easier if I just push a commit with the changes to the manifest based on your suggested changes to this PR.

rdf/rdf12/rdf-semantics/manifest.ttl
```diff
@@ -102,7 +102,7 @@ trs:annotated-asserted a mf:PositiveEntailmentTest;
   test:approval test:NotClassified .
 
 trs:annotation a mf:PositiveEntailmentTest;
-  rdfs:comment "Annotation are about the reifying triple.";
+  rdfs:comment "Annotations are about the reifying triple.  This is about shorthand expansion, and is not really a semantics test.";
```
TallTed, line 105: I think this (and similar) is in error. `the reifying triple` is the triple with `rdfs:reifies` predicate. `the reified triple` is the one about which an annotation is written.
```suggestion
  rdfs:comment "Annotations are about the reified triple.  This is about shorthand expansion, and is not really a semantics test.";
```

PR #202, opened 2025-06-27
Add missing prefixes for Graph Store Protocol Test

Qup42:
The expected response for the Graph Store Protocol test `GET of PUT - Initial state` was missing the prefix definitions. This PR adds these.


afs:
Thank you!

PR #203, opened 2025-06-27
Indirect Graph Identification Graph IRI is absolute

Qup42:
If [Graph Store Protocol Indirect Graph Identification](https://www.w3.org/TR/2013/REC-sparql11-http-rdf-update-20130321/#indirect-graph-identification) is used, then the query string must be an absolute IRI. The Graph Store Protocol test `GET of PUT - Initial state` had a relative IRI which is now fixed.


Tpt (approved):
Thank you!

It might be great to add an other test with a relative IRI to cover <q>the server MUST respond with a 400 Bad Request if it is not [an absolute IRI].</q>

afs (commented):
Bug fix. No IPR impact.

afs (approved):

Qup42:
> It might be great to add an other test with a relative IRI to cover the server MUST respond with a 400 Bad Request if it is not [an absolute IRI].

Thats a good idea. @tpt Can this just be done as a test without approval for now or would it have to go through the responsible WG like the other approved tests?

afs:
It can be added to this PR.

PR #205, opened 2025-06-28
Add test for entailment of literal type

niklasl:
Adds test for second semantic condition of [RDF Interpretations](https://www.w3.org/TR/rdf12-semantics/#rdf_d_interpretations) (based on [example pattern](https://www.w3.org/TR/rdf12-semantics/#rdf_entailment_patterns)).

### DIFF
--- a/rdf/rdf12/rdf-semantics/index.html
+++ b/rdf/rdf12/rdf-semantics/index.html
@@ -527,6 +527,31 @@
             </dd>
           </dl>
         </dd>
+        <dt id='literal-type'>
+          <a class='testlink' href='#literal-type'>
+            literal-type:
+          </a>
+          <span about='../rdf-semantics#literal-type' property='mf:name'>literal-type</span>
+        </dt>
+        <dd inlist='true' property='mf:entry' resource='../rdf-semantics#literal-type' typeof='mf:PositiveEntailmentTest'>
+          <div property='rdfs:comment'>
+            <p>Literals denote instances of their datatype.</p>
+          </div>
+          <dl class='test-detail'>
+            <dt>type</dt>
+            <dd>mf:PositiveEntailmentTest</dd>
+            <dt>approval</dt>
+            <dd property='mf:approval' resource=''></dd>
+            <dt>action</dt>
+            <dd>
+              <a href='canonical-literal-control.ttl' property='mf:action'>canonical-literal-control.ttl</a>
+            </dd>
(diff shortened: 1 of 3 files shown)


niklasl:
> Should this be in the RDF 1.1 tests?

Good point; I believe they should. There is currently no rdf11/rdf-semantics directory; but I can set that up in this PR. Are there other tests that should be moved there? (Presumably in a separate PR.)

pfps:
The float and double tests, at least.

niklasl:
> > > Should this be in the RDF 1.1 tests?
> > 
> > 
> > Good point; I believe they should. There is currently no rdf11/rdf-semantics directory; but I can set that up in this PR. Are there other tests that should be moved there? (Presumably in a separate PR.)
> 
> In 1.1, it was rdf-mt, so rdf/rdf11/rdf-mt contains all the 1.1 semantics tests.

Ah, of course; thanks! I'll move this there (unless I've missed such a case there...).

gkellogg:
It would be good to get to just a single manifest, and integrate any -az tests that are appropriate.

in the past, working groups marked approved vs proposed tests. My view now, is that the PR process satisfies this, and any tests added via a PR are effectively approved.

afs:
@niklasl  - this has a merge conflict.

Other than that, I think this PR is mergeable, and the discussions on other matters taken to a separate issue/PR.

niklasl:
I've addressed the conflict.

But the github report generation action fails for some reason (perhaps due to the PR being from my fork). This might be OK to merge anyway; generating the reports locally again yields no change; which seems in order as they where already regenerated (with this addition being the only modification). Unless the generation itself should be backed out (if the action expects to take care of it)...

afs:
> (perhaps due to the PR being from my fork)

Not sure but I think we can ignore it.
We'll get some more data when it is merged which AFAICT will generate report files.

It looks like the error (unexpected 128 result code) is from the Ruby environment. Any help for this would be appreciated.

(The "add to WG project" does fail for PRs from a separate repo.)

afs:
The error for action "Generating Reports" is now:

https://github.com/w3c/rdf-tests/actions/runs/18156713919/job/51678172547

> remote: error: GH006: Protected branch update failed for refs/heads/main.        
> remote: 
> remote: - Changes must be made through a pull request.        

so (1) not related to this PR (2) might be an environmental factor.

PR #225, opened 2025-10-01
Adds test to prevent repeated variables in VALUES

Tpt:
Test for https://github.com/w3c/sparql-query/pull/294

### DIFF
--- a/rdf/rdf12/rdf-semantics/index.html
+++ b/rdf/rdf12/rdf-semantics/index.html
@@ -541,7 +541,17 @@
             <dt>type</dt>
             <dd>mf:PositiveEntailmentTest</dd>
             <dt>approval</dt>
-            <dd property='mf:approval' resource=''></dd>
+            <dd property='mf:approval' resource=''>none</dd>
+            <dt>entailmentRegime</dt>
+            <dd property='mf:entailmentRegime'></dd>
+            <dt>recognizedDatatypes</dt>
+            <dd>
+              <code inlist='true' property='mf:recognizedDatatypes' resource='xsd:integer'>xsd:integer</code>
+            </dd>
+            <dt>unrecognizedDatatypes</dt>
+            <dd>
+              <code property='mf:unrecognizedDatatypes' resource='rdf:nil'>none</code>
+            </dd>
             <dt>action</dt>
             <dd>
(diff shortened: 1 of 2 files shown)


rdf/rdf12/rdf-semantics/index.html
```diff
@@ -541,7 +541,17 @@ <h2>
             <dt>type</dt>
             <dd>mf:PositiveEntailmentTest</dd>
             <dt>approval</dt>
-            <dd property='mf:approval' resource=''></dd>
+            <dd property='mf:approval' resource=''>none</dd>
```
afs, line 544: This may become unnecessary if another PR is merged. This looks like it is the report generator catching up.

sparql/sparql12/expression/duplicated-values-variable.rq
```diff
@@ -0,0 +1 @@
+SELECT * WHERE { VALUES (?a ?a) { (1 1) } }
```
afs, line 1: I don't think this is the right place.

1. It isn't an expression.
1. All syntax tests go in a directory name of the form `syntax-*`, separate from eval tests.

Suggestion: start a new directory: `sparql/sparql12/syntax` for both positive `syntax-*.rq` tests and `syntax-bad-*.rq` for syntax tests that do not warrant their own directory.

Discussion: should this be in `sparql11`?

Tpt: Indeed, a syntax dataset would be great

> Discussion: should this be in sparql11?

imho it's undefined behavior in SPARQL 1.1 and is only defined in SPARQL 1.2 so an implementation doing something else would be still compliant with SPARQL 1.1. This is why I put this test in SPARQL 1.2.

afs: > This is why I put this test in SPARQL 1.2.

OK

sparql/sparql12/syntax/duplicated-values-variable.rq
```diff
@@ -0,0 +1 @@
+SELECT * WHERE { VALUES (?a ?a) { (1 1) } }
```
Tpt: Moved into a new syntax dir

PR #240, opened 2025-10-24
SPARQL 1.2: add an evaluation test with annotation syntax

Tpt:
The annotation syntax was not covered in `ObjectListPath` evaluation


rubensworks (approved):
Looks correct, the test also succeeds on my end.

PR #244, opened 2025-10-28
SPARQL 1.0: ORDER BY is applied before projecting variables

Tpt:
Test access of a non-projected variable in ORDER BY


afs:
There are changes to `index.html` in `sparql12/` in this PR.

They look right but I thought I'd mention it in case it s sign of not being in-sync.

Tpt:
@afs It's because we just merged some PRs from forks for which the html creation is not working

PR #253, opened 2025-12-05
Fix GSP `POST - existing graph` and `GET of POST - existing graph` test mixup/omission

Qup42:
Fix the description of the tests `POST - existing graph` and `GET of POST - existing graph` in the manifest.

- `POST - existing graph` had no comment describing the test case.
- `GET of POST - existing graph` contained the test case for `POST - existing graph`.

The specifications from <https://www.w3.org/2009/sparql/docs/tests/data-sparql11/http-rdf-update/>, but are also what I'd expect from the test names. As far as I can tell both tests have been [approved in 2012](https://www.w3.org/2009/sparql/meeting/2012-11-27#resolution_3), but I don't know why they got lost/mixed up.


Tpt (approved):
Thank you!

rubensworks (approved):
Looks good, but I don't have a way to test these myself, so I can't guarantee that they work.

PR #254, opened 2025-12-10
materialize JSON-LD versions of the manifests

pchampin:
note that these versions are generated internally to generate the HTML versions. This change simply writes the results in a file.
Some people might find this more convenient to process than the TTL files?

I don't feel strongly about this,
I ended up doing this while getting myself familiarized with the automation process.


Tpt (approved):
Nice addition for implementations that might want to run some tests without a proper Turtle parser

pchampin:
> Rather than JSON-LD in the tree (FWIW they used to be produced), I think it is better to produce the jsonld files as the GH-pages are created. Then there is one version of the truth.

I sympathize with that, and to mitigate the potential issues, I added a "comment" at the top of each JSON-LD file, explaining that they are generated and should not be edited.

Arguably, the whole generation process could be restricted to the page generation, rather than materialized in the repo. Or do you consider that the HTML pages are different from the JSON-LD files?

afs:
> we should probably also release zips containing all the files,

Github provides download as zip.

https://github.com/w3c/rdf-tests/archive/refs/heads/main.zip

afs:
> Some people might prefer that

Please substantiate "some". We simply can not accommodate every possibility we can think of. Why can't we assume that users can convert using local or online tools if necessary for them? This repo contains Turtle and N-Triples tests. They are likely to have a TTL parser.

The tradeoff is having one version of the truth.

And it does get out of step. GH actions fail, then the newest tests are in the TTL and not the JSON-LD yet they are the focus at the time.

afs:
By "release zip", do you mean the built state at a point in time with a version number?

Tpt:
I have rebased the MR. It now generates the JSON-LD files in the same gh-pages branch as the HTML files. I also apply to the JSON-LD files the preprocessing done for HTML templating. This way we "dogfood" them when generating the HTML pages.

@afs @pchampin @rubensworks new review welcome!

afs (approved):

rubensworks (approved):

PR #260, opened 2025-12-16
Beginning of manifest validations using SHACL

Tpt:
Only validates some fields of RDF tests for now


rubensworks (approved):
This is really useful, thanks for the effort @Tpt.

When validation fails, how will it be reported within the logs?
Does it show the filename of the file that was invalid (with line number)?

afs (approved):

Tpt:
> How about putting the shapes in a different place to /ns/ such as /build/ or /.github/shapes/?

@afs We can but I wanted to avoid yet an other directory and `/.github/shapes/` is a bit burried for something that is not specific to the CI. What about putting the shapes into a `manifest-shape.ttl` file in the root directory?

Tpt:
@afs Done. aa30368f43d7fc66750a76953143cf960b62b605

I am not sure to see the problem with naming the shape file "manifest". The main RDF node of the files listing tests uses the `mf:Manifest` class.

ns/test-manifest.ttl
```diff
@@ -77,7 +77,7 @@
     # rdfs:range   ?? ;
     .	
 
-:result rdf:type rdf:Property ;
+:status rdf:type rdf:Property ;
```
Tpt, line 81: copy of #358 to make shapes pass

rubensworks (approved):
Looks right to me!

shapes/shapes.ttl
```diff
@@ -0,0 +1,368 @@
...
+PREFIX sd: <http://www.w3.org/ns/sparql-service-description#>
+PREFIX sh: <http://www.w3.org/ns/shacl#>
+PREFIX ut: <http://www.w3.org/2009/sparql/tests/test-update#>
+PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
+
+[] a sh:NodeShape ;
+    sh:targetClass mf:Manifest ;
+    sh:property
+        [
+            sh:path mf:include ;
```
afs, line 18: SHACL 1.2 has constraint checking for lists.

Tpt: Good call! Added a beginning. Sadly pyshacl does not support it yet

PR #265, opened 2025-12-18
Update triple term <<()>> syntax tests

afs:
Align to  w3c/sparql-query#332.


afs:
>These tests will fail on generalized RDF implementations. Not sure if it's a big deal though.

FWIW `TRIPLE` can be extended - there, it's an eval error and can be extended.

PR #268, opened 2025-12-26
SPARQL 1.1: REPLACE: cover regex options

Tpt:
Use only 'i', REGEX covers regex options in details


afs (approved):
Approved with a minor suggestion (not required) to emphasis the change.

kasei:
Oops. I waited too long. After seeing this this morning, I was wondering if it should cover more than just the `i` flag. Maybe I'll try to create a new issue/PR.

PR #273, opened 2025-12-29
SPARQL 1.2: rdf:dirLangString with DATATYPE, CONCAT and CONTAINS

Tpt:
Covers argument compatibility


afs:
There are more cases such as 
`CONCAT("abc"@en--ltr, "xyz"@en--rtl)` (-> xsd:string)
`CONCAT("abc"@en--ltr, "xyz"@en)` (-> xsd:string, and not `@en`)
`CONTAINS("abc"@en--ltr, "a"@en)` (->incompatible)

I'll sort out a collection of these.

PR #291, opened 2026-02-19
moving some semantics tests from rdf12 to rdf11

### LINKED ISSUE #207: location of semantics tests for float and double (RDF 1.1 vs RDF 1.2)
I proposed some tests for float and double.  They were added to the rdf12 area but are correct for rdf11 as well.  Should these tests be moved to the rdf11 area?  

Should all new tests that are correct for rdf11 be put on the rdf11 area as well?

pchampin:
because these tests are actually good for RDF 1.1 Semantics as well.

fix #207

### DIFF
--- a/rdf/rdf12/rdf-semantics/manifest.ttl
+++ b/rdf/rdf12/rdf-semantics/manifest.ttl
@@ -227,16 +227,6 @@ trs:triple-terms-no-spurious a mf:NegativeEntailmentTest;
   mf:unrecognizedDatatypes ();
   rdft:approval rdft:NotClassified .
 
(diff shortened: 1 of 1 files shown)


rdf/rdf12/rdf-semantics/manifest.ttl
niklasl: The file `rdf/rdf12/rdf-semantics/canonical-literal-control.ttl` was moved to `rdf/rdf11/rdf-mt/datatypes/literal-type1.ttl` in this PR, but it still used here. I suggest copying it instead (as I should have done from the start).

franconi (approved):

Issue #284, opened 2026-02-05
SPARQL manifest generation missing links


afs:
Example: https://w3c.github.io/rdf-tests/sparql/sparql11/functions/index.html

only has the result files.

<img width="501" height="304" alt="Image" src="https://github.com/user-attachments/assets/2210baae-d9a4-4d4a-99c5-b39c9104ea14" />

PR #295, opened 2026-02-25
2026 02 improve html generation

pchampin:
fix some issue in the HTML rendering of the Semantics and SPARQL test suite

also remove some redundancy in the frames and templates used for the generation,
and add a `<link>`  from all HTML manifests to their original TTL version


pchampin:
now #284 should be completely fixed

rubensworks (approved):

PR #296, opened 2026-02-25
More testing of triple terms in RDF/XML

afs:
This PR:

1 - adds a test for 
```xml
    <ex:prop rdf:parseType="Triple">
      <ex:t rdf:about="http://example/s"/>
    </ex:prop>
```
where the triple term is given in a way other than `rdf:Description`.

2  - It moves negative syntax tests to `rdf12-xml-tt-bad*`

3 - It updates the results for `rdf12-xml-tt-01`. The results were empty (no triples). I don't understand that.

```xml
<?xml version="1.0"?>
<rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
            xmlns:ex="http://example.org/stuff/1.0/"
            xml:base="http://example.org/triples/">
  <rdf:Description rdf:about="http://example.org/">
    <ex:prop rdf:parseType="Triple">
      <rdf:Description rdf:about="http://example.org/stuff/1.0/s">
        <ex:p rdf:resource="http://example.org/stuff/1.0/o" />
      </rdf:Description>
    </ex:prop>
  </rdf:Description>
</rdf:RDF>
```

I get one triple with a triple term: (reformatted because of long lines)

```
<http://example.org/> 
    <http://example.org/stuff/1.0/prop> 
        <<( <http://example.org/stuff/1.0/s> <http://example.org/stuff/1.0/p> <http://example.org/stuff/1.0/o> )>> .
```

### DIFF
--- a/rdf/rdf12/rdf-xml/eval/manifest.ttl
+++ b/rdf/rdf12/rdf-xml/eval/manifest.ttl
@@ -40,7 +40,8 @@ trs:manifest  rdf:type mf:Manifest ;
         trs:rdf12-xml-tt-05
         trs:rdf12-xml-tt-06
         trs:rdf12-xml-tt-07
-        trs:rdf12-xml-tt-08
+        trs:rdf12-xml-tt-bad-01
+        trs:rdf12-xml-tt-bad-02
 
         trs:rdf12-xml-an-01
         trs:rdf12-xml-an-02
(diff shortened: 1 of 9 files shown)


Tpt:
> rdf:version="1.2"  isn't in other tests e.g. rdf12-xml-an-01 either but the output includes triple terms.

Great point! So, your test change make sense.

> or parseType="Triple", it's an error in RDF v1.1.

Thanks! I got it wrong in my parser. It might be nice to add a test for this error.

Tpt (approved):

afs:
Yesterday, the WG voted to progress Turtle, N-Quads and TriG to horizontal review.

The window on RDF/XML is still open.

PR #298, opened 2026-02-27
Dockerfile and bash script to generate HTML files locally

pchampin:
NB: generation of HTML files are normally managed by github actions. This script is provided for the sole purpose of testing the generation, but the generated files must not be committed to the repository.

maybe this should not be merged to the public repo,
but at least some of the maintainers may find it useful to get it in their working copy.

### DIFF
--- a/local-gen/.gitignore
+++ b/local-gen/.gitignore
@@ -1,0 +1,2 @@
+Gemfile
+Gemfile.lock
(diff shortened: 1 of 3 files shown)


rubensworks (approved):
To me, this seems useful to the repo at least :-)

PR #301, opened 2026-02-27
RDF/XML: add a test to make sure the parser fails on unknown rdf:parseType value

Tpt:


afs:
I have every sympathy with the intent here!

But
(https://www.w3.org/TR/rdf11-xml/#parseTypeOtherPropertyElt)

> All rdf:parseType attribute values other than the strings "Resource", "Literal" or "Collection" are treated as if the value was "Literal". This production matches and acts as if production [parseTypeLiteralPropertyElt](https://www.w3.org/TR/rdf11-xml/#parseTypeLiteralPropertyElt) was matched. No extra triples are generated for other rdf:parseType values.

whatever the scope of "No extra triples are generated" is:
"No extra triples" has two readings: 
1. No triples; match production but no action (see @rubensworks https://github.com/w3c/rdf-tests/pull/296#issuecomment-3972923318)
2. No triples other than generating an `rdf:XMLLiteral` (ARP[^1])

The idea of parsing and producing either nothing, or something other than intended, seems at odds with accurate publishing of data. c

Now we have `rdf:parseType="Triple"` which will in RDF 1.1 should pass parsing, I'm not sure what the best thing to do is.

Maybe it's an errata, arguing that the old behaviour does not accurately round trip the input.

[^1]: ARP is Jena's RDF/XML parser from around the time of the RDF 1.0 WG. It is not the one users are using.
It generates a `rdf:XMLLiteral` (case 2 above), with a warning.
Nowadays, the RRX parser makes it a parse error.

Jena does have some hard-wired fixed-ups, e.g. literal" vs "Literal", "Statements" because CIM uses that for "Literal".

PR #309, opened 2026-03-18
RDF/XML test for reification and annotation together

afs:
A test that has both RDF 1.1 reification and RDF 1.2 annotation on the same triple.


rubensworks (approved):
Looks good, I get the same result.

PR #322, opened 2026-04-13
Fix incorrect Content-Length in test case text description

Qup42:
Fix an error in the `Content-Length` of the text descriptions of one test case introduced in #312. @kasei


Tpt (approved):

kasei (approved):

kasei:
Looks good. I'll note, however, that I think we should just remove these long `rdfs:comment` strings that have full HTTP traces in them (now that #312 added better modeling of this data).

Qup42:
The modelling looks promising. It'll take some time to adapt our conformance test runner but I expect that the changes will make it much easier to use the tests. I am also looking forward to seeing this expanded to other tests with raw HTTP traces, like the GSP tests.

PR #177, opened 2025-05-30
RDF/XML 1.2 features

gkellogg:
Tests Initial Text Direction, Triple Terms, and Annotations.
Also adds additional 1.1 reification tests.

PR #325, opened 2026-04-20
Align xml-canon to previous behavior

afs:
The original test in [rdf11/rdf-xml/xml-canon/test001.rdf](https://github.com/w3c/rdf-tests/blob/main/rdf/rdf11/rdf-xml/xml-canon/)
for `rdf:XMLLiteral` changed and it seems to have changed RDF/XML behaviour.

https://github.com/w3c/rdf-tests/blob/main/rdf/rdf11/rdf-xml/xml-canon/
https://github.com/w3c/rdf-tests/pull/177 (added test002)
https://github.com/w3c/rdf-tests/pull/206 (changes test001.nt)

The RDF 1.1 tests are included in the RDF 1.2 tst manifest so an implementation that passed at RDF 1.1 will not pass when they try RDF 1.2.

There is wider issue: https://github.com/w3c/rdf-xml/issues/97
This PR is only for reverting `rdf11/rdf/xml-canon` behavior.

### DIFF
--- a/rdf/rdf11/rdf-xml/xml-canon/test001.nt
+++ b/rdf/rdf11/rdf-xml/xml-canon/test001.nt
@@ -13,4 +13,4 @@
 # $Id: test001.nt,v 1.1 2003/08/18 10:09:29 jgrant Exp $
 #
 #####################################################################
-<http://www.example.org/a> <http://example.org/prop> "<br xmlns:rdf=\"http://www.w3.org/1999/02/22-rdf-syntax-ns#\" xmlns:eg=\"http://example.org/\"></br>"^^<http://www.w3.org/1999/02/22-rdf-syntax-ns#XMLLiteral> .
+<http://www.example.org/a> <http://example.org/prop> "<br></br>"^^<http://www.w3.org/1999/02/22-rdf-syntax-ns#XMLLiteral> .
--- a/rdf/rdf11/rdf-xml/xml-canon/test002.nt
+++ b/rdf/rdf11/rdf-xml/xml-canon/test002.nt
@@ -11,8 +11,8 @@
 # Canonicalization of XMLLiterals with reification.
 #
 #####################################################################
-<http://www.example.org/a> <http://example.org/prop> "<br xmlns:rdf=\"http://www.w3.org/1999/02/22-rdf-syntax-ns#\" xmlns:eg=\"http://example.org/\"></br>"^^<http://www.w3.org/1999/02/22-rdf-syntax-ns#XMLLiteral> .
+<http://www.example.org/a> <http://example.org/prop> "<br></br>"^^<http://www.w3.org/1999/02/22-rdf-syntax-ns#XMLLiteral> .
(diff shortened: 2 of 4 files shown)


rubensworks:
I'm the one who was responsible for the `xml-canon-test001` change: https://github.com/w3c/rdf-tests/pull/206
However, this was done purely based on my understanding of the 1.1 spec (so not 1.2-related), and comparing the output with @gkellogg's implementation.
My interpretation could have been wrong though.

afs:
Did you have to change your implementation to get that result?

I came across this while catching up on various changes and assumed Jena was wrong but trying converter.zazuko and Sophia, I also get `"<br></br>"`. I hunted around for an original test001.nt (RDF 1.0 generation) and it does not have unused namespaces. 

https://github.com/w3c/rdf-xml/issues/97 is the result of trying to determine the right answer ... which I didn't manage from the RDF 1.1 XML text.

Tpt:
> trying converter.zazuko and Sophia, I also get "<br></br>"

Sophia last release is using an old version of my RDF/XML parser that was not including any namespaces when parsing `rdf:parseType="Literal"` literals. After #206 I changed it to always include all namespaces.

rubensworks:
> Did you have to change your implementation to get that result?

Yes, I did: https://github.com/rdfjs/rdfxml-streaming-parser.js/commit/98d6ab5a5d5d14f551cc0427ec85ebee3ed73acd
I usually document the reasoning behind changes quite well, but apparently not this time, so I don't remember the precise reason for this change...
I suspect I ran into some inconsistencies across the different spec tests (possibly the other canon test), but I'm not sure atm.
I'll need to do some deeper analysis on why the change was done.

> but trying converter.zazuko and Sophia

zazuko's tool also uses my parser internally, but probably still using the older version.

afs:
RDF 1.1/ RDF 1.2 possibly introduced a change from RDF 1.0. 

1.1/1.2 say "uses `fn:serialize`", but that has a one-argument and a two-argument form. The second argument being a number of parameters to the serialization process. The one argument form is "no parameters" and I think "no settting" leads to no specific default.

One such parameter is [`undeclare-prefixes`](https://www.w3.org/TR/xslt-xquery-serialization-30/#XML_OMIT-XML-DECLARATION), values "yes" or "no".  The default if unset seems to be "undefined" from ...

> Additional namespace [nodes](https://www.w3.org/TR/xslt-xquery-serialization-30/#dt-node) MAY be present in the [reconstructed tree](https://www.w3.org/TR/xslt-xquery-serialization-30/#reconstructed-tree) if the serialization process did not undeclare one or more namespaces, as described in [5.1.8 XML Output Method: the undeclare-prefixes Parameter](https://www.w3.org/TR/xslt-xquery-serialization-30/#xml-undeclare-NS), ...

The default for indent/whitespace is also "MAY".

I haven't found a connection between [Canonical XML](https://www.w3.org/TR/xml-c14n11/) and [XSLT and XQuery Serialization](https://www.w3.org/TR/xslt-xquery-serialization-30).

rubensworks (approved):
I just updated my implementation based on the discussions around this PR, and I can confirm that I get the same output now.

(apologies for the long delay)

Tpt (approved):

PR #326, opened 2026-04-26
Include older SPARQL tests in newer manifests

### LINKED ISSUE #307: Explicitly add tests from older SPARQL standards to the manifest of newer standards
It would be very convenient and useful, if the manifests for the SPARQL 1.1 and 1.2 tests contained a reference to compatible tests from older standards like 1.1 and 1.0. This makes it easier for people implementing the SPARQL 1.2 and 1.1 standards to discover tests that were originally written for older standards but are still valid in newer versions. As I was told in https://github.com/w3c/rdf-tests/issues/305#issuecomment-4026342812, there are some tests that are incompatible with newer versions, but most of them should be fine most of the time.
Thank you.

erdemonal:
Fixes #307

### DIFF
--- a/sparql/sparql11/manifest-all.ttl
+++ b/sparql/sparql11/manifest-all.ttl
@@ -6,6 +6,7 @@
 <> a mf:Manifest ;
 	rdfs:label "SPARQL 1.1 tests" ;
 	mf:include (
+		<../sparql10/manifest.ttl>
 		<add/manifest.ttl>
 		<aggregates/manifest.ttl>
 		<basic-update/manifest.ttl>
@@ -55,6 +56,7 @@
 
 <http://www.w3.org/TR/sparql11-query/> rdfs:label "SPARQL 1.1 Query Language" ;
 	mf:conformanceRequirement (
+		<../sparql10/manifest.ttl>
 		<aggregates/manifest.ttl>
 		<bind/manifest.ttl>
 		<bindings/manifest.ttl>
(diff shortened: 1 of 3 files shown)


rubensworks:
I believe the tests are intentionally separated like this to be able to run test manifests for each SPARQL version independently.

afs:
Having worked with the RDF tests with included style, I prefer them separate. It means the 1.2 test responses reflect the state just of 1.2 features. 

As we gather implementation reports this will be more convenient. Once the implementation report phase is over, then maybe combine them.

lisp:
if the issue is that a test is superseded or obsoleted, it would be better that have a protocol whereby a test with the same identifier in a later manifest can modify the approval status of an the earlier test to be withdrawn or obsoleted.  
i see nothing in https://www.w3.org/2001/sw/DataAccess/tests/README.html in that regard.  
that could be changed.

afs:
> if the issue is that a test is superseded or obsoleted, it would be better that have a protocol whereby a test with the same identifier in a later manifest can modify the approval status of an the earlier test to be withdrawn or obsoleted. i see nothing in https://www.w3.org/2001/sw/DataAccess/tests/README.html in that regard. that could be changed.

No - it's not the issue.

An SPARQL 1.2 must pass SPARQL 1.0 and 1.1 tests - indeed the tests here wre updated for RDF 1.1 changes around `xsd:string`.

1.2 tests are tests for features not in 1.1 or, sometimes, clarification tests. They are not replacements.

lisp:
if it is the case that, once 1.1 test had been corrected, any "SPARQL 1.2 must pass SPARQL 1.0 and 1.1 tests", then there is no reason to make this change.

kasei:
> As we gather implementation reports this will be more convenient. Once the implementation report phase is over, then maybe combine them.

I think a better approach might be to have a "top"-level sparql/manifest.ttl that just `mf:include`d the version-specific manifests. This would leave intact the ability to run the version-specific tests if desired.

PR #327, opened 2026-05-02
Remove SPARQL Graph Store Protocol test gsp:put__mismatched_payload.

kasei:
While this was approved by the SPARQL 1.1 WG, it does not seem to make any sense. The test name mentions "mismatched payload", but the payload in this test matches the content type in the header, and is the exact same request as in test gsp:put__graph_already_in_store (which expects success).

Have discussed this with @Tpt and @afs and agreed that removing the test was the best path forward.

### DIFF
--- a/sparql/sparql11/http-rdf-update/manifest.ttl
+++ b/sparql/sparql11/http-rdf-update/manifest.ttl
@@ -15,7 +15,6 @@
                  gsp:get_of_put__graph_already_in_store
                  gsp:put__default_graph
                  gsp:get_of_put__default_graph
-                 gsp:put__mismatched_payload
                  gsp:delete__existing_graph
                  gsp:get_of_delete__existing_graph
                  gsp:delete__nonexistent_graph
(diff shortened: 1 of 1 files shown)


afs (approved):

lisp:
the protocol tests have already a test, :bad_query_wrong_media_type, when covers the case where the media type specifies one which is not suitable for the request. this test could be corrected to embody a mismatch and moved to the protocol tests. it would be similar to the :bad_query_missing_form_type test, but with an explicit conflict.

kasei:
> this test could be corrected to embody a mismatch and moved to the protocol tests.

I don't understand "moved to the protocol tests." This is a GSP test, not a Protocol test. I agree we could probably add a new test that tries to follow the implication of the "mismatched payload." Such a test might be like this one but with a changed `Content-Type` header. This PR chooses simply to remove the test as it seems obviously wrong. We could add a new test as a follow-up.

lisp:
> corrected to embody a mismatch and moved to the protocol tests

does mean

> add a new test [to the protocol tests] that tries to follow the implication of the "mismatched payload."

kasei:
> > corrected to embody a mismatch and moved to the protocol tests
> 
> does mean
> 
> > add a new test [to the protocol tests] that tries to follow the implication of the "mismatched payload."

I think I'm still misunderstanding something here. I wrote the second quote above, but the parts you've added in brackets is not what I said nor intended. This is a GSP test. It has nothing to do with Protocol, which is an entirely different spec. I was suggesting that we could follow up by adding a new GSP test in place of what is being removed here.

Tpt (approved):

rubensworks (approved):

lisp:
i propose that, as a protocol level test, it would be best combined with other protocol tests.

afs:
The W3C short name for the spec is [sparql11-http-rdf-update/](https://www.w3.org/TR/sparql11-http-rdf-update/). 

And, importantly, it is a separate manifest. Not all systems support GSP.

lisp:
> I was suggesting that we could follow up by adding a new GSP test in place of what is being removed here.

if the intent is to keep "sparql graph store protocol" tests in collection distinct from "sparql protocol" tests, then this would be appropriate.  
the test collections' names would be less ambiguous if they were revised to be more specific.

PR #333, opened 2026-05-16
new test checking boolean keywords with unusual case

pchampin:
it is not just a syntax test, because we must also make sure that they are not returned as ill-typed literal (e.g. "TRUE"^^xsd:boolean)

### DIFF
--- a/sparql/sparql12/expression/case-insensitive-booleans.srj
+++ b/sparql/sparql12/expression/case-insensitive-booleans.srj
@@ -2,7 +2,7 @@
     "head": {
         "vars": [
             "t",
-            "f",
+            "f"
         ]
     },
     "results": {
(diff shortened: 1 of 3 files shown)


rubensworks (approved):

Tpt (approved):
Thanks! I would put the test in the sparql-10 folder, it's covering a SPARQL 1.0 feature

sparql/sparql12/expression/case-insensitive-booleans.srj
```diff
@@ -0,0 +1,24 @@
+{
+    "head": {
+        "vars": [
+            "t",
+            "f",
```
afs:
```suggestion
            "f"
```
This is JSON, not javascript!
https://datatracker.ietf.org/doc/html/rfc8259#section-5

afs (changes requested):

afs (approved):

kasei (approved):

PR #324, opened 2026-04-16
manifest with rdf action model, revised with additional tests as per …

lisp:
…requirements.md

this follows on to the version which i had described at the beginning of february.
the reaction then was to prefer that the test notation be uniformly rdf, rather than a combination of rdf and javascript.
in the interim gregory revised the existing 1.1 version to incorporate gregg's approach.

this version extends gregory's version.
it introduces a requirements document which was generated from the draft recommendation.
most prominent among that are test which exercise the version information.
it then add several tests which follow form those requirements.

PR #343, opened 2026-05-24
Update SPARQL Protocol test :update_base_uri to allow simple verification

kasei:
The previous test used an `mf:expectation` assertion to describe the test expectation in prose. The new test is changed to use an `ASK` query, with a true response indicating success.

This is based on the change by @lisp in #324 (pulled out into its own PR) with one change: Part of the verification filter is changed from `?o != <test>` to `STR(?o) != "test"`. I think this is more aligned with the original intention and description of the test, and avoids the risk that both the update and the verification query use the same base IRI, causing the non-equality test to fail.


rubensworks (approved):

Tpt (approved):

afs (approved):

lisp:
i will pull a new version before i work on the various issues which have been noted in the pr.

PR #345, opened 2026-05-25
Update GSP tests to improve modeling and fix existing bugs.

### LINKED ISSUE #252: `GET of POST - create new graph` and `GET of POST - after noop` access graph at wrong location?
`GET of POST - create new graph` and `GET of POST - after noop` access a location that can be incorrect. Indirect graph identification should be used instead of direct graph identification for these cases.
 
- The location of `POST - create new graph` points to the graph store itself. The data is inserted into a new graph created by the store. The IRI of the new graph is returned in the `Location` header.
- `GET of POST - create new graph` and `GET of POST - after noop` retrieve the data from the newly created graph

Assume that the GSP endpoint of the graph store is available at `http://example.com/http-graph-store/`. As long as it is new there are no restrictions on what form the IRI of the new graph must satisfy by [the standard](https://www.w3.org/TR/sparql-graph-store-protocol/#http-post). Let's assume it is `http://example.com/internal-graphs/78123`. `$NEWPATH$` contains the IRI of this newly created graph as returned by the `Location` header. The problem now is that tests request `GET $NEWPATH$ HTTP/1.1` which can be a path that is completely unrelated to GSP depending on the graph IRI. The current target only works if the created graph IRI is suitable for direct graph identification.

This could be fixed by using indirect graph identification for these tests: `GET $GRAPHSTORE$?graph=$NEWPATH$ HTTP/1.1`.

kasei:
The previous text-based modeling included many bugs and was essentially impossible to use as the source of truth for automated testing. This change uses the modeling introduced in #312 and attempts to preserve the intent of the existing tests, with several differences:

1. Each test assumes an empty initial dataset.
2. Many tests define a sequence of several request-response pairs that depend on the state changes of previous requests in the same test.
3. There is a new mf:expectedLocation property used to make the Location response header value returned after the creation of a new graph to be made available as a template variable in subsequent requests within the same test.

With this new modeling, the tests in http-rdf-update/ are marked as deprecated and removed from manifest-all.ttl, with the new tests being added to that manifest and existing in the new graph-store-protocol/ directory.

Closes #252

### DIFF
--- a/sparql/sparql11/http-rdf-update/manifest.ttl
+++ b/sparql/sparql11/http-rdf-update/manifest.ttl
@@ -2,6 +2,9 @@
 @prefix dawg: <http://www.w3.org/2001/sw/DataAccess/tests/test-dawg#> .
 @prefix gsp: <http://www.w3.org/2009/sparql/docs/tests/data-sparql11/http-rdf-update/manifest#> .
 @prefix mf: <http://www.w3.org/2001/sw/DataAccess/tests/test-manifest#> .
+@prefix cnt:    <http://www.w3.org/2011/content#> .
+@prefix ht:     <http://www.w3.org/2011/http#> .
+@prefix hts:    <http://www.w3.org/2011/http-statusCodes#> .
 @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
 @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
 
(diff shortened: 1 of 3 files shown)


rubensworks (approved):

sparql/sparql11/http-rdf-update/manifest.ttl
```diff
@@ -2,6 +2,9 @@
 @prefix dawg: <http://www.w3.org/2001/sw/DataAccess/tests/test-dawg#> .
 @prefix gsp: <http://www.w3.org/2009/sparql/docs/tests/data-sparql11/http-rdf-update/manifest#> .
 @prefix mf: <http://www.w3.org/2001/sw/DataAccess/tests/test-manifest#> .
+@prefix cnt:    <http://www.w3.org/2011/content#> .
+@prefix ht:     <http://www.w3.org/2011/http#> .
+@prefix hts:    <http://www.w3.org/2011/http-statusCodes#> .
```
Tpt: Unused prefixes
```suggestion
```

Tpt (approved):

afs (approved):
Jena has not have a GSP test runner yet but that is no reason not to merge these tests.

PR #346, opened 2026-05-30
Codepoint escapes updates

afs:
Test changes for https://github.com/w3c/sparql-query/issues/382

Bonus: in the same directory, bad syntax tests for the use of surrogates (pairs and broken).


Tpt (approved):
Oxigraph passes the test.
Might be nice to add tests to cover how the decoding is done (e.g. `"\u005C\t"` is the same as `"\\\t"` and not `"\\t"`).
Glad to do it in a follow-up

kasei (approved):

rubensworks (approved):
For reference, our parser does not conform to these tests yet.

PR #347, opened 2026-06-05
Split GSP tests for varying support of graph identification/creation.

kasei:
All tests now use either direct or indirect (or default) graph identification, and are classified using `mf:requires` assertions to differentiate direct and indirect.

For tests that use direct or indirect graph identification, test variants now exist for both of the methods.

Also, the one test for having the Graph Store create a new graph by POSTing to the endpoint URL is now also marked with an `mf:requires` assertion.


afs (approved):
Feel free to merge this it is as one manifest and split into two later in another PR.

Tpt (approved):
Thank you!

> EDIT: Sorry, mistake with my test runner. I think oxigraph passes all tests except for the ones that use multipart/form-data (the post_get_post_get_direct_* tests).

Yes, multipart/form-data is not implemented yet in Oxigraph

afs (approved):

PR #349, opened 2026-06-09
Add a GSP test for percent-decoding of indirect-identified graphs.

kasei:
As requested in comments of w3c/sparql-graph-store-protocol#30

### DIFF
--- a/sparql/sparql11/graph-store-protocol/manifest-indirect.ttl
+++ b/sparql/sparql11/graph-store-protocol/manifest-indirect.ttl
@@ -17,6 +17,7 @@
                  gsp:post_get_new_graph
                  gsp:head_existing_indirect
                  gsp:head_non_existing_indirect
+                 gsp:put_get_uri_pct_encoded_indirect
                ) .
 
 gsp:put_get_repeat_indirect rdf:type mf:GraphStoreProtocolTest ;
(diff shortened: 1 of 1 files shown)


rubensworks (approved):

Tpt (approved):
Thank you!

sparql/sparql11/graph-store-protocol/manifest-indirect.ttl
```diff
@@ -623,3 +624,79 @@ gsp:head_non_existing_indirect rdf:type mf:GraphStoreProtocolTest ;
...
 
+gsp:put_get_uri_pct_encoded_indirect rdf:type mf:GraphStoreProtocolTest ;
+    mf:name    "PUT/GET (indirect graph identification; using varying percent-encoding for the same graph IRI)" ;
+	mf:requires mf:IndirectGraphIdentification ;
+    mf:action [
+        a ht:Connection ;
+        ht:connectionAuthority "www.example" ;
+        ht:requests ([
+                a ht:Request ;
+                ht:absolutePath "/gsp?graph=http%3A%2F%2Fwww.example%2Fgsp%2Fperson%2F1.ttl" ;
```
Tpt, line 635:
```suggestion
                ht:absolutePath "/gsp?graph=http%3A%2F%2Fwww.example%2Fgsp%2Fperson%2F%2531.ttl" ;
```

sparql/sparql11/graph-store-protocol/manifest-indirect.ttl
```diff
@@ -623,3 +624,79 @@ gsp:head_non_existing_indirect rdf:type mf:GraphStoreProtocolTest ;
...
+                ht:resp [
+                    a ht:Response ;
+                    mf:expectedStatus hts:Created
+                ]
+            ]
+            [
+                a ht:Request ;
+                ht:methodName "GET" ;
+                ht:httpVersion "1.1" ;
+                ht:absolutePath "/gsp?graph=http://www.example/gsp/person/1.ttl" ;
```
Tpt, line 667:
```suggestion
                ht:absolutePath "/gsp?graph=http://www.example/gsp/person/%31.ttl" ;
```

Tpt (commented):
idea to check double escapes

sparql/sparql11/graph-store-protocol/manifest-indirect.ttl
afs: Isn't that double encoding?  
The `absolutePath` is after encoding.

Line 667 uses `%31`.

Tpt: Yes! Because we want to check the implementation decode only once (to `%31`, hence line 667) but not twice (to `1`).

But maybe I am missing something

sparql/sparql11/graph-store-protocol/manifest-indirect.ttl
kasei: I think I'd prefer if any/all of these suggestions were new tests. I'm not sure this last suggestion is adding any real validation (if it is, let's make it explicit with a new test).

sparql/sparql11/graph-store-protocol/manifest-indirect.ttl
kasei: Testing with double encoding (one decoded by the protocol, and one left as-is in the graph name) seems like a fine idea, but:

* I'd prefer it in a different test
* I'm not sure how you'd validate the actual graph name without using a SPARQL query (since GSP only has the one path to indirectly identifying encoded graph names).

Tpt: We can test for double encoding by checking that we do not fetch the same graph with double encoding and single encoding? I am going to submit a follow up when we meege this one

afs (approved):

PR #350, opened 2026-06-10
SPARQL: Do not require empty named graph support to pass the query tests

Tpt:
Change two tests introduced recently to tests GRAPH behavior to not rely on empty named graphs

### DIFF
--- a/sparql/sparql11/aggregates/manifest.ttl
+++ b/sparql/sparql11/aggregates/manifest.ttl
@@ -425,7 +425,7 @@
 
 :agg-empty-group-count-graph  rdf:type mf:QueryEvaluationTest ;
 	mf:name "COUNT: no GROUP BY inside of GRAPH" ;
-    rdfs:comment "counting no results without grouping always returns a single result per named graph." ;
+    rdfs:comment "counting without grouping always returns a single result per named graph." ;
     mf:action
          [ qt:query     <agg-empty-group-count-graph.rq> ;
            qt:data      <empty.ttl> ;
--- a/sparql/sparql11/aggregates/manifest.ttl
+++ b/sparql/sparql11/aggregates/manifest.ttl
@@ -429,8 +429,8 @@
     mf:action
          [ qt:query     <agg-empty-group-count-graph.rq> ;
            qt:data      <empty.ttl> ;
-           qt:graphData <empty.ttl> ;
-           qt:graphData <singleton.ttl> ] ;
+           qt:graphData <singleton.ttl> ;
+           qt:graphData <pair.ttl> ] ;
     mf:result  <agg-empty-group-count-graph.ttl>
     .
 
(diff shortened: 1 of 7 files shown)


sparql/sparql11/aggregates/manifest.ttl
```diff
@@ -429,8 +429,8 @@
     mf:action
          [ qt:query     <agg-empty-group-count-graph.rq> ;
            qt:data      <empty.ttl> ;
-           qt:graphData <empty.ttl> ;
-           qt:graphData <singleton.ttl> ] ;
+           qt:graphData <singleton.ttl> ;
+           qt:graphData <pair.ttl> ] ;
```
kasei, line 433: Doesn't removing `empty.ttl` move the test away from the behavior being tested ("counting no results without grouping always returns a single result per named graph")?

I see the issue you're trying to address, but I think the fix here is actually reducing the test coverage. Maybe we just need another `mf:requires` assertion?

sparql/sparql11/aggregates/manifest.ttl
Tpt:
```suggestion
    rdfs:comment "counting without grouping always returns a single result per named graph." ;
```

sparql/sparql11/aggregates/manifest.ttl
Tpt: That's a good point. It indeed reduces a bit coverage. A new `mf:requires` assertion might make sense but in this case it would be nice to duplicate the test to have also coverage for endpoints not supporting empty named graph.

I have updated the MR to fix the description of the test.

I am fine with either option but would love to get @afs opinion.

rubensworks (approved):
This looks correct at first glance, but I can't validate it, as our implementation doesn't pass this test (old and new) yet.

sparql/sparql11/aggregates/manifest.ttl
```diff
@@ -425,12 +425,12 @@
...
 	mf:name "COUNT: no GROUP BY inside of GRAPH" ;
-    rdfs:comment "counting no results without grouping always returns a single result per named graph." ;
+    rdfs:comment "counting without grouping always returns a single result per named graph." ;
     mf:action
          [ qt:query     <agg-empty-group-count-graph.rq> ;
            qt:data      <empty.ttl> ;
-           qt:graphData <empty.ttl> ;
-           qt:graphData <singleton.ttl> ] ;
+           qt:graphData <singleton.ttl> ;
+           qt:graphData <pair.ttl> ] ;
```
afs, line 433: How about having a pattern that does not match in one case so the result is still 0 but it's not because the graph is empty?

sparql/sparql11/aggregates/manifest.ttl
Tpt: Done with @afs approach: a filter create an empty graph. Does it works for you?

sparql/sparql11/aggregates/manifest.ttl
Tpt: Thank you! Done!

afs (approved):

kasei (approved):

PR #351, opened 2026-06-12
Add a GSP test for single but not double percent-decoding of indirect-identified graphs

Tpt:


sparql/sparql11/graph-store-protocol/manifest-indirect.ttl
```diff
@@ -700,3 +700,94 @@ gsp:put_get_uri_pct_encoded_indirect rdf:type mf:GraphStoreProtocolTest ;
     ] ;
     .
 
+gsp:put_get_uri_pct_encoded_twice rdf:type mf:GraphStoreProtocolTest ;
```
kasei, line 704: Test needs to be added to the `mf:entries` list at the top of the file.

sparql/sparql11/graph-store-protocol/manifest-indirect.ttl
```diff
@@ -700,3 +700,94 @@ gsp:put_get_uri_pct_encoded_indirect rdf:type mf:GraphStoreProtocolTest ;
...
+    mf:name    "PUT/GET/GET (indirect graph identification; checking that we decode percent-encoding only once (for the request URL query) and not twice)" ;
```
kasei: nit: I would prefer avoiding the double-parenthesized text here. The `name` value here could be simplified, or the extra context would be added to the test with `rdfs:comment`.

sparql/sparql11/graph-store-protocol/manifest-indirect.ttl
```diff
@@ -700,3 +700,94 @@ gsp:put_get_uri_pct_encoded_indirect rdf:type mf:GraphStoreProtocolTest ;
...
+                                a v:VCard;
+                                v:fn "John Doe"
+                            ].
+                    """
+                ] ;
+                ht:httpVersion "1.1" ;
+                ht:methodName "PUT" ;
+                ht:resp [
+                    a ht:Response ;
+                    mf:expectedStatus hts:Created
```
kasei, line 737: I think this is right per the spec, but I notice that in some of the PUT tests above, we expect `hts:OK, hts:Created, hts:NoContent`. I suspect I added that because I found implementations that weren't returning `hts:Created` (and carelessly followed the implementations, not the spec). I don't think anything needs to change here, but lets' follow-up with fixing the other cases (and maybe finding any implementations that don't return 201 and let the developers know).

sparql/sparql11/graph-store-protocol/manifest-indirect.ttl
```diff
@@ -700,3 +700,94 @@ gsp:put_get_uri_pct_encoded_indirect rdf:type mf:GraphStoreProtocolTest ;
...
+                ht:absolutePath "/gsp?graph=http://www.example/gsp/person/%2531.ttl" ;
+                ht:headers ([
+                        a ht:RequestHeader ;
+                        ht:fieldName "accept" ;
+                        ht:fieldValue "text/turtle"
+                    ]
+                ) ;
+                ht:resp [
+                    a ht:Response ;
+                    mf:expectedStatus hts:NotFound ;
```
kasei: I'm not sure about this. The spec doesn't actually have any relevant spec text on response codes for `GET`, but we have to be careful here about systems that can't differentiate non-existent graphs and empty graphs. I *think* this means that a conformant system should either return 404 *or* a RDF serialization with zero triples, but we don't have a way (yet) to represent that expectation in this manifest. As an alternative, I think we could restructure this test to do two PUTs as initialization (one to the single-encoded graph, and one to the double-encoded graph), with different content in each, and then validate that you get back the (differing) expected RDF content for each of the two corresponding GETs.

I think we also probably need to improve the normative language in the GET section of the spec regarding what is required.

sparql/sparql11/graph-store-protocol/manifest-indirect.ttl
TallTed:
```suggestion
    mf:name    "PUT/GET/GET (indirect graph identification; checking that we decode percent-encoding only once  — for the request URL query — and not twice)" ;
```

sparql/sparql11/graph-store-protocol/manifest-indirect.ttl
Tpt: This is a great point. Thank you! I have rewritten the test as you suggested.

+1 to improve the GET section of the spec.

kasei (approved):

afs (approved):

Issue #143, opened 2024-09-25
SPARQL syntax test classes in test manifest vocabulary include version in IRI (`11`)


rubensworks:
The [test manifest vocabulary](http://www.w3.org/2001/sw/DataAccess/tests/test-manifest) introduces the following:

```
:PositiveSyntaxTest11 rdf:type rdfs:Class ;
      rdfs:label "Positive Syntax Test for SPARQL1.1 Query" ;
      rdfs:comment "A type of test specifically for syntax testing of new features in the SPARQL1.1 Query Language. Syntax tests are not required to have an associated result, only an action." .

:PositiveUpdateSyntaxTest11 rdf:type rdfs:Class ;
      rdfs:label "Positive Syntax Test for SPARQL1.1 Update" ;
      rdfs:comment "A type of test specifically for syntax testing of SPARQL1.1 Update. Syntax tests are not required to have an associated result, only an action." .

:NegativeSyntaxTest11 rdf:type rdfs:Class ;
      rdfs:label "Negative Syntax Test for SPARQL1.1 Query" ;
      rdfs:comment "A type of test specifically for syntax testing of new features in the SPARQL1.1 Query Language. Syntax tests are not required to have an associated result, only an action. Negative syntax tests are tests of which the result should be a parser error." .

:NegativeUpdateSyntaxTest11 rdf:type rdfs:Class ;
      rdfs:label "Negative Syntax Test for SPARQL1.1 Update" ;
      rdfs:comment "A type of test specifically for syntax testing of SPARQL1.1 Update. Syntax tests are not required to have an associated result, only an action. Negative syntax tests are tests of which the result should be a parser error." .
```

This means that the test classes are hardcoded to SPARQL 1.1, which is inconsistent with the test classes for RDF syntaxes (`PositiveSyntaxTest` and `NegativeSyntaxTest`), which do not include such a version label.

In order to add new syntax tests for SPARQL 1.2, we can either

1. broaden the usage of the SPARQL test classes, and keep using classes such as `PositiveSyntaxTest11` for SPARQL 1.2, or
2. introduce `PositiveSyntaxTestSparql` and variants into the vocabulary, and deprecate `PositiveSyntaxTest11` and variants.

Option 2 seems like the cleanest option to me.

afs:
For completeness, there is a "3":
3. Add a `PositiveSyntaxTest12` and related names.

Note:
[...]

PR #354, opened 2026-06-19
SPARQL 1.1: Remove the Test11  type variants

Tpt:
We have not included new type variants for 1.2 and consensus in #143 seems to suggest we could remove them for 1.1 too


afs:
Now #356 is on `main`, can this PR restart? via manifest level entries presumably, not per test.

I came across the `NegativeSyntaxTest11`/`NegativeSyntaxTest` difference after resyncing which wasn't a problem because ARQ has a "test adaption layer" to convert to one set of test types.

PR #356, opened 2026-06-20
SPARQL: Set mf:specVersion on manifests and use IRIs

Tpt:


rubensworks (approved):

afs (approved):

sparql/sparql10/algebra/manifest.ttl
```diff
@@ -4,9 +4,11 @@
 @prefix mf:     <http://www.w3.org/2001/sw/DataAccess/tests/test-manifest#> .
 @prefix qt:     <http://www.w3.org/2001/sw/DataAccess/tests/test-query#> .
 @prefix dawgt:   <http://www.w3.org/2001/sw/DataAccess/tests/test-dawg#> .
+@prefix sparql: <http://www.w3.org/ns/sparql#> .
 
 <>  rdf:type mf:Manifest ;
     rdfs:label "Algebra" ;
+    mf:specVersion sparql:version-1.0 ;
```
kasei, line 11: This might be bordering on the absurd, but are there any concerns about using Turtle qnames that wouldn't have been syntactically valid when SPARQL 1.0 was standardized (I don't think the dot was valid at that point)?

afs: No concern here.

SPARQL 1.0 (2008) precedes the Turtle team submission (2011).
Turtle didn't exist as a standard until RDF 1.1 (2014) and at that point it had internal DOT in local names.

RDF 1.1 changed the semantics of strings so the `"Algebra"` would be a problem!

afs:
Can we get this merged so downstream implementers have fixed target for their work?

While we're not at formal implementation phase yet, earlier is better.

Tpt:
@afs feel free to sqash and merge if you think it's time

kasei (approved):

PR #357, opened 2026-06-20
SPARQL: Tag *Test11 tests as deprecated and make them aliases of regular tests

Tpt:


ns/test-manifest.ttl
```diff
@@ -114,11 +115,15 @@
 
 :PositiveSyntaxTest11 rdf:type rdfs:Class ;
       rdfs:label "Positive Syntax Test for SPARQL 1.1 Query" ;
-      rdfs:comment "A type of test specifically for syntax testing of new features in the SPARQL 1.1 Query Language. Syntax tests are not required to have an associated result, only an action." .
+      rdfs:comment "A type of test specifically for syntax testing of new features in the SPARQL 1.1 Query Language. Syntax tests are not required to have an associated result, only an action." ;
+      owl:deprecated true ;
+      rdfs:subClassOf :PositiveSyntaxTest .
```
afs, line 120: This for documentation only?

If RDFS is active, then there are multiple test types.

`PositiveSyntaxTest11` "is" the pair (`PositiveSyntaxTest`, "sparql-1.1").

Either there is an assumption that no inference is performed to get the unique test type
or the most specific type should be chosen (which sort of defeats the migration)
or a custom inference that PositiveSyntaxTest11 also gives the specVersion.

Something that needs to be documenting in the instructions
Alternativey,  use a comment to say that PositiveSyntaxTest11 implies specVersion. 

Writing this -- I favor the comment approach; documentation somewhere else is easily overlooked.

((
This seems to come up regularly. 
Having synonyms is not equivalent to having "owl:sameAs"/"rdfs:subClassOf"/... as far as cardinality is concerned.
))

ns/test-manifest.ttl
```diff
@@ -114,11 +115,15 @@
...
 
 :PositiveUpdateSyntaxTest11 rdf:type rdfs:Class ;
       rdfs:label "Positive Syntax Test for SPARQL 1.1 Update" ;
-      rdfs:comment "A type of test specifically for syntax testing of SPARQL 1.1 Update. Syntax tests are not required to have an associated result, only an action." .
+      rdfs:comment "A type of test specifically for syntax testing of SPARQL 1.1 Update. Syntax tests are not required to have an associated result, only an action." ;
```
afs: Unrelated fix!

```suggestion
      rdfs:comment "A type of test specifically for syntax testing of SPARQL 1.1 Update. Syntax tests do not have an associated result, only an action." ;
```

ns/test-manifest.ttl
Tpt: > This for documentation only?

Mostly, indeed.

> If RDFS is active, then there are multiple test types.

Indeed. Note this is already happening for the parsing tests (`TestNTriplesPositiveSyntax` is subclass of `TestSyntax` that is itself subclass of `Test`) so I hope runners using RDFS are also supporting multiple types.

>  Having synonyms is not equivalent to having "owl:sameAs"/"rdfs:subClassOf"/... as far as cardinality is concerned.

Yes!

> Writing this -- I favor the comment approach; documentation somewhere else is easily overlooked.

Agreed. Done: 8796e8ea177c1e52814ce7e69dc9790f0e17e45d

kasei (approved):

rubensworks (approved):

afs (approved):

PR #362, opened 2026-06-27
Clean up superficial test output and manifest inconsistencies

drobilla:
Some minor cleanups towards more consistent/canonical test outputs, along with some superficial manifest tidying.

The larger motivation here is to reduce the disparity between the syntax emitted by implementations and the given expected outputs. Ideally this should be eliminated entirely, but that would require more significant (but still just superficial) changes, so this is a very small and hopefully uncontroversial step towards that, which at least reduces the number of changes that developers of implementations that check test outputs exactly (like mine) need to manually re-fix when updating the test suite.

### DIFF
--- a/rdf/rdf11/rdf-trig/IRI-resolution-01.nq
+++ b/rdf/rdf11/rdf-trig/IRI-resolution-01.nq
@@ -1,42 +1,42 @@
-<urn:ex:s001> <urn:ex:p> <g:h>.
-<urn:ex:s002> <urn:ex:p> <http://a/bb/ccc/g>.
-<urn:ex:s003> <urn:ex:p> <http://a/bb/ccc/g>.
-<urn:ex:s004> <urn:ex:p> <http://a/bb/ccc/g/>.
-<urn:ex:s005> <urn:ex:p> <http://a/g>.
-<urn:ex:s006> <urn:ex:p> <http://g>.
-<urn:ex:s007> <urn:ex:p> <http://a/bb/ccc/d;p?y>.
-<urn:ex:s008> <urn:ex:p> <http://a/bb/ccc/g?y>.
-<urn:ex:s009> <urn:ex:p> <http://a/bb/ccc/d;p?q#s>.
-<urn:ex:s010> <urn:ex:p> <http://a/bb/ccc/g#s>.
-<urn:ex:s011> <urn:ex:p> <http://a/bb/ccc/g?y#s>.
-<urn:ex:s012> <urn:ex:p> <http://a/bb/ccc/;x>.
-<urn:ex:s013> <urn:ex:p> <http://a/bb/ccc/g;x>.
-<urn:ex:s014> <urn:ex:p> <http://a/bb/ccc/g;x?y#s>.
-<urn:ex:s015> <urn:ex:p> <http://a/bb/ccc/d;p?q>.
-<urn:ex:s016> <urn:ex:p> <http://a/bb/ccc/>.
-<urn:ex:s017> <urn:ex:p> <http://a/bb/ccc/>.
-<urn:ex:s018> <urn:ex:p> <http://a/bb/>.
-<urn:ex:s019> <urn:ex:p> <http://a/bb/>.
-<urn:ex:s020> <urn:ex:p> <http://a/bb/g>.
-<urn:ex:s021> <urn:ex:p> <http://a/>.
-<urn:ex:s022> <urn:ex:p> <http://a/>.
-<urn:ex:s023> <urn:ex:p> <http://a/g>.
+<urn:ex:s001> <urn:ex:p> <g:h> .
+<urn:ex:s002> <urn:ex:p> <http://a/bb/ccc/g> .
+<urn:ex:s003> <urn:ex:p> <http://a/bb/ccc/g> .
+<urn:ex:s004> <urn:ex:p> <http://a/bb/ccc/g/> .
+<urn:ex:s005> <urn:ex:p> <http://a/g> .
+<urn:ex:s006> <urn:ex:p> <http://g> .
+<urn:ex:s007> <urn:ex:p> <http://a/bb/ccc/d;p?y> .
+<urn:ex:s008> <urn:ex:p> <http://a/bb/ccc/g?y> .
+<urn:ex:s009> <urn:ex:p> <http://a/bb/ccc/d;p?q#s> .
+<urn:ex:s010> <urn:ex:p> <http://a/bb/ccc/g#s> .
+<urn:ex:s011> <urn:ex:p> <http://a/bb/ccc/g?y#s> .
+<urn:ex:s012> <urn:ex:p> <http://a/bb/ccc/;x> .
+<urn:ex:s013> <urn:ex:p> <http://a/bb/ccc/g;x> .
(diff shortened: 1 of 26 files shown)


kasei (approved):
Looks good to me. Thanks for the attention on this.

Tpt (approved):

afs (commented):
I don't see any problems with the changes but the description suggests other changes which would be a problem.

In an NT/NQ file, there is no guarantee of order. (I think we do ensure a line can be "parsed" with a regex.)

Trying to get Turtle list-parsing to be order-defined would need a spec change even for streaming support e.g. the order of `rdf:first` and `rdf:next`.

Parsing the file and comparing graphs/datasets with blank node isomorphism is necessary.

Fortunately, the tests generate very small results and bnode-isomorphism can be done simply.

> inconsistent generated blank node names

This should not matter because blank node names are scoped to the file, and are not suitable, in general, for the internal system id which needs to preserve uniqueness across files and parser run. If we chose one style, then we are favoring one system over another.

RDF Canonicalization relies on its own algorithm for blank node labels to make sign-able files. Canonical NQ does what it can in support of RDF Canonicalization but RDF Canonicalization needs its own control as well.

rdf/rdf11/rdf-trig/manifest.ttl
```diff
@@ -2677,127 +2677,127 @@
 # tests from David Robillard
 # http://www.w3.org/2011/rdf-wg/wiki/Turtle_Candidate_Recommendation_Comments#c21
 <#trig-syntax-bad-blank-label-dot-end>
-	rdf:type rdft:TestTrigNegativeSyntax ;
-	rdfs:comment "Blank node label must not end in dot" ;
-	mf:name "trig-syntax-bad-blank-label-dot-end" ;
-        rdft:approval rdft:Approved ;
-	mf:action <trig-syntax-bad-blank-label-dot-end.trig> .
+   rdf:type rdft:TestTrigNegativeSyntax ;
```
afs: Getting rid if tabs is good - same display for every one.

Did you mean 4 spaces, not 3?

The mf:entries list is 4 - mostly, some are 6.

The turtle manifest uses 4 space.

rdf/rdf11/rdf-trig/IRI-resolution-01.nq
afs, line 1: Variants of this content are in other places (e.g. *.trig) and all have no space before the DOT.

rdf/rdf11/rdf-trig/IRI-resolution-01.nq
```diff
@@ -21,7 +21,6 @@
 <urn:ex:s021> <urn:ex:p> <http://a/> .
 <urn:ex:s022> <urn:ex:p> <http://a/> .
 <urn:ex:s023> <urn:ex:p> <http://a/g> .
-
```
afs, line 24: The blank line reflects the TriG file (which seems to be the point where ../../ runs out) so I think it is better to leave it in.

rdf/rdf11/rdf-turtle/turtle-syntax-bad-number-dot-in-anon.ttl
```diff
@@ -1,6 +1,3 @@
 @prefix : <http://www.w3.org/2013/TurtleTests/> .
 
-:s
```
afs, line 3: Maybe the point of the test is that NLs don't matter? Is this covered somewhere else?

rubensworks (approved):

rdf/rdf11/rdf-trig/manifest.ttl
drobilla: I was shooting for minimal changes overall, so went with the style of the surrounding pieces, but I'd be happy to do everything in 4 spaces if everyone is fine with that.

rdf/rdf11/rdf-trig/IRI-resolution-01.nq
drobilla: I changed only the outputs on purpose, since this is the part that needs to be normalized to support testing parsers/serializers in isolation with simple text comparison. Style diversity in inputs is good if anything (it covers more).

rdf/rdf11/rdf-trig/IRI-resolution-01.nq
drobilla: No other outputs have flair like this, and no implementation will produce it.

rdf/rdf11/rdf-turtle/turtle-syntax-bad-number-dot-in-anon.ttl
drobilla: This one is originally from myself/serd. It's just that way because all my documents are formatted in the same way, the "27." is the point of the test. Since it's an input, this one isn't important, I just noticed it while going through this stuff and figured I'd clean it up so it didn't look strange (and potentially mislead about the point of the test).

As an aside, if there's interest in more coverage, I have many more tests than are in the official suite. It would take a bit of work to separate the ones that are generally interesting from the ones that are specifically tailored to cover my implementation, but probably not too much. They are broadly categorized already though, see https://github.com/drobilla/serd/tree/main/test/extra

afs: If the others are fine with changing output files, then OK.

It is changing input files that I'm not so sure of.

* [trig-syntax-bad-number-dot-in-anon.trig](https://github.com/w3c/rdf-tests/blob/main/rdf/rdf11/rdf-trig/trig-syntax-bad-number-dot-in-anon.trig)
* [turtle-syntax-bad-number-dot-in-anon.ttl](https://github.com/w3c/rdf-tests/blob/main/rdf/rdf11/rdf-turtle/turtle-syntax-bad-number-dot-in-anon.ttl)

(yes, I know the layout isn't the point of the test.)

Is this something you have to maintain in the serd your copy of the tests?

drobilla: There's no concrete reason these matter, no (the whitespace before the "." is significant, this is a Turtle 1.1ism that's more annoying to parse, the rest is just coincidence). I just cleaned them up because I happened to notice and they're originally my "fault". It's unlikely anything would be affected, but I agree not modifying the inputs is a sensible precaution, so I'll drop this part.

rdf/rdf11/rdf-trig/IRI-resolution-01.nq
drobilla: "changed only the outputs" was a slight lie, resolved by dropping 968ad51

rdf/rdf11/rdf-turtle/turtle-syntax-bad-number-dot-in-anon.ttl
drobilla: Dropped 968ad51

rdf/rdf11/rdf-trig/manifest.ttl
drobilla: Turns out many places, including the entire RDF 1.2 suite, uses two spaces, and I don't want to cause conflicts on almost every single line in an active working area for no good reason, so here's just the (formerly inconsistent) RDF 1.1 ones 2004679

afs (approved):

PR #366, opened 2026-08-13
Add collection point for RDF 1.2 implementation reports

### LINKED ISSUE #187: Confirm organisation of implementation reports
From w3c/rdf-concepts#258.

In the [RDF 1.1 test cases document](https://www.w3.org/TR/2014/NOTE-rdf11-testcases-20140225/), each of the concrete test suites has it's own "/reports" area.

"rdf-tests" is a CG whereas there will be a "maintenance and new features" WG.

Might it be better in each of the document github repos (as is done for, say, JSON-LD)?

rubensworks:
This closes https://github.com/w3c/rdf-star-wg/issues/187

This is mostly based upon the way JSON-LD collects implementation reports, and makes use of @gkellogg's earl reporter.
I don't have much Ruby experience, so I made the minimal changes to the setup in JSON-LD to make it work.
There's currently a dummy report for my rdf-parse library to show that it works.

If we agree with this approach, we can add a similar setup for the SPARQL tests.

Related to w3c/rdf-star-wg#142, w3c/rdf-star-wg#187, #359


rdf/rdf12/template.haml
Tpt, line 1: Is it intentional? I guess it might break the HTML entry page for rdf12 tests

Tpt (commented):
Thank you for this! I think the CI is failing because the jobs are trying to act on the introduced `manifests.ttl` file, expecting it to be a test manifest files.

The auto-generated files are big, instead of pushing them to the main branch, what about doing the same things as the manifest html/jsonld renderings and do it in the CI?

rdf/rdf12/template.haml
rubensworks: Was not intentional; will re-add.

rubensworks:
@Tpt I made the necessary changes to make it generate in the CI.

Tpt (approved):
@rubensworks amazing! Thanks!

afs (approved):

rubensworks:
For reference, the RDF 1.2 reports are now available here: https://w3c.github.io/rdf-tests/rdf/rdf12/reports/
(maybe we should add a link from the index page?)

I will soon provide a similar setup for SPARQL.

PR #370, opened 2026-08-25
Add qt:graphData to the SPARQL 1.0 dataset test manifest.ttl

noahvsb:
According to @rubensworks's [comment on my PR](https://github.com/comunica/comunica/pull/1746#issuecomment-5341205191) the dataset spec tests for SPARQL 1.0 are incorrectly defined. The fix is to add `qt:graphData <...>` for the graphs appearing in `FROM (NAMED)` clauses. So that's exactly what I did in this PR.


afs:
> the dataset spec tests for SPARQL 1.0 are incorrectly defined. 

The original assumption is that the FROM/FROM NAMED are URLs that are read by HTTP - in this set of tests, files in teh same location as the query (file: or http:).

These changes make a different assumption, naming picking graphs out of an underlying dataset - a common choice but not mandated by the spec.

&nbsp;

I get a test failure on 12b.

Look like it is because the default graph data isn't set but in the query has:

```
FROM <data-g1-dup.ttl>
FROM <data-g2-dup.ttl>
FROM <data-g3-dup.ttl>
FROM <data-g4-dup.ttl>
```

&nbsp;

Queries `dataset-12.rq`, `dataset-09.rq`, `dataset-10.rq`, aren't used and can be removed.

&nbsp;

`dataset-04` has a typo! `qt:grapthData`
So, for ARQ, no dataset is setup, and it reads the local file and passes.

&nbsp;

@noahvsb - how does comunica pass 04 and 12b?

rubensworks:
> The original assumption is that the FROM/FROM NAMED are URLs that are read by HTTP - in this set of tests, files in teh same location as the query (file: or http:).

The reason why @noahvsb added the changes here, is because we could not find anywhere in the spec that these HTTP lookups were required, so we assumed it was a mistake.
@afs Do you see where this is mentioned in the spec? Or is this an implicit assumption? (and if the latter, must these tests be considered valid or invalid?)

afs:
> Do you see where this is mentioned in the spec? 

https://www.w3.org/TR/sparql10-query/#specifyingDataset
"obtained from representations of the resources identified by the given IRIs"

"obtained" is intentionally neutral as to how it happens. Its declarative. There are graphs, there is mapping of IRI to graph. HTTP satisfies that as does look up in another (base) dataset. Neither is required.

The driver (IIRC) was that "data on the web" was a thing back then.

rubensworks:
> https://www.w3.org/TR/sparql10-query/#specifyingDataset
"obtained from representations of the resources identified by the given IRIs"

I see that the wording has not changed since then, as is still the same in 1.2.

I see that most SPARQL implementations do not dereference anything when using FROM/NAMED, and interpret this part of the spec as a way to specify the select default/named graphs.

@afs Instead of modifying these tests as done in this PR, do you think it would make sense to annotate them with an flag to indicate that they require a dereference feature/mode? (similar to the `sd:DereferencesURIs` mode described in [SD](https://w3c.github.io/sparql-service-description/spec/#sd-dereferencesuris))
This way, it becomes clearer that there are different ways of interpreting the spec regarding this point.

afs:
> @afs Instead of modifying these tests as done in this PR, do you think it would make sense to annotate them with an flag to indicate that they require a dereference feature/mode? (similar to the `sd:DereferencesURIs` mode described in [SD](https://w3c.github.io/sparql-service-description/spec/#sd-dereferencesuris))
> This way, it becomes clearer that there are different ways of interpreting the spec regarding this point.

I agree it is a common way. 

Nowadays security issues are much more prominent and SSRF (server call out) is not seen as desirable so favor the "local" interpretation is reasonable IMO.

Modifying the manifest as per the PR wil be fine.

It is the default behavior in Jena esp Fuseki. 
The command line CLI sparql tool can deference URLs (give it a query, but no dataset).

Documenting can be either an `mf:requires`/`sd:DereferencesURIs false` - or because this is unique to this test suite - a comment in the manifests (it is probably why it is called `open-world`).

rubensworks:
@noahvsb Could you modify the PR to not use `graphData`, but instead use `mf:requires` as discussed above?

noahvsb:
A change in https://www.w3.org/2001/sw/DataAccess/tests/README.html is also needed.

afs:
> @noahvsb Could you modify the PR to not use `graphData`, but instead use `mf:requires` as discussed above?

For me, using `qt:graphData` is fine. It can be seen as generally keeping the tests up-to-date, favoring the more popular, and also no SSRF.

Like we did for RDF 1.1 plain literal -> `xsd:string`.

rubensworks:
@Tpt what's your view on this? I'm fine with either option (graphdata or requires).

Tpt:
> @Tpt what's your view on this? I'm fine with either option (graphdata or requires).

I am also fine with both option. I have a slight preference for graphData because it will make the logic in Oxigraph testsuite runner slightly easier but I do not care strongly. I agree with @afs that changing the tests is fine.

rubensworks (approved):

sparql/sparql10/dataset/manifest.ttl
```diff
@@ -30,7 +30,8 @@
     dawgt:approval dawgt:Approved ;
     dawgt:approvedBy <http://lists.w3.org/Archives/Public/public-rdf-dawg/2007JulSep/att-0047/31-dawg-minutes> ;
     mf:action
-            [ qt:query  <dataset-01.rq> ] ;
+            [ qt:query       <dataset-01.rq> ;
+              qt:graphData   <data-g1.ttl> ] ;
```
Tpt, line 34: It might be nice to add other `gt:graphData` to this tests and some others. In the current state of the manifest I think a SPARQL evaluator that just ignore FROM NAMED clauses would pass some of the tests


```suggestion
              qt:graphData   <data-g1.ttl> ;
              qt:graphData   <data-g2.ttl> ] ;
```

Tpt (approved):
Thanks! Oxigraph works with these tests. Just a small suggestion to make them better

afs (approved):

sparql/sparql10/dataset/manifest.ttl
afs: > this tests and some others.

Separate PR?

PR #371, opened 2026-08-25
Add implementation reports for n3.js and rdfxml-streaming-parser.js

rubensworks:
Adds reports for https://github.com/rdfjs/N3.js/ and https://github.com/rdfjs/rdfxml-streaming-parser.js/


Tpt (approved):

rubensworks:
(fixed the merge conflicts)

afs:
Something to sort out later ...

If it is one system, would we be accumulating the various files and presenting them as one implementation report/experience?

Each report says:
```
 doap:implements <https://www.w3.org/TR/turtle/>, <https://www.w3.org/TR/trig/>, <https://www.w3.org/TR/n-triples/>, <https://www.w3.org/TR/n-quads/>;
```

afs (approved):

rubensworks:
> If it is one system, would we be accumulating the various files and presenting them as one implementation report/experience?

Based on the current process, it shouldn't matter too much, as all triples from all reports are collected in one place, and based on that the HTML is generated.
This way, we should be aligned with tools that produce either one large report per tool, or separate ones per tool-spec.

afs:
So we want the subject node to be an IRI, not a blank node.

rubensworks:
I don't thing it necessarily indexes on the subject node, but on the object on `foaf:primaryTopic`.
(This is all using @gkellogg's code, so I'm also not super familiar with all of the details; I just wired things together here)

Issue #305, opened 2026-03-07
Proposing new SPARQL 1.1 compliance tests


RobinTF:
This is a follow-up issue to my comment in https://github.com/w3c/rdf-tests/issues/243#issuecomment-3694708681.
Here's a list of compliance tests I came up with, for review. If they don't contain any mistakes I plan on eventually creating a PR for each individual one:

### Query
```sparql
BASE <http://example.com/xxx/abc>
SELECT (<> AS ?res0) (<def> AS ?res1) (<#ghi> AS ?res2) (<./jkl> AS ?res3) (<../mno> AS ?res4) (</pqr> AS ?res5) {}
```

### Expected Result
```
?res0	?res1	?res2	?res3	?res4	?res5
<http://example.com/xxx/abc>	<http://example.com/xxx/def>	<http://example.com/xxx/abc#ghi>	<http://example.com/xxx/jkl>	<http://example.com/mno>	<http://example.com/pqr>
```

### Query
```sparql
SELECT (""@en = ""@EN AS ?res0) (LANG(""@en) = LANG(""@EN) AS ?res1) (STRLANG("", "en") = STRLANG("", "EN") AS ?res2) {}
```

### Expected Result
```
?res0	?res1	?res2
true	true	true
```

#### `graph1.ttl`
```ttl
@prefix : <http://example.com/> .
:s :p :o
:s1 :p :o1
```

#### `graph2.ttl`
```ttl
@prefix : <http://example.com/> .
:s :p :o
:s2 :p :o2
```

### Query
```sparql
PREFIX : <http://example.com/>
SELECT ?s ?o WHERE {
  GRAPH ?g {
    GRAPH ?g {
      ?s :p ?o
    }
  }
}
ORDER BY ?s ?o
```

### Expected Result
```
?s	?o
<http://example.com/s> <http://example.com/o>
<http://example.com/s> <http://example.com/o>
<http://example.com/s1> <http://example.com/o1>
<http://example.com/s2> <http://example.com/o2>
```

#### `graph1.ttl`
```ttl
@prefix : <http://example.com/> .
:s :p :o
:s1 :p :o1
```

#### `graph2.ttl`
```ttl
@prefix : <http://example.com/> .
:s :p :o
:s2 :p :o2
```

### Query
```sparql
PREFIX : <http://example.com/>
SELECT ?s ?o WHERE {
  GRAPH ?g1 {
    GRAPH ?g0 {
      ?s :p ?o
    }
  }
}
[...]

PR #374, opened 2026-09-23
SPARQL 1.0: add nested `GRAPH` tests

RobinTF:
This is a PR that implements part of the tests proposed in #305. Adds four tests for nested `GRAPH` patterns combining graph variables and fixed graph IRIs.


Tpt (approved):

kasei:
I think these tests look fine, but unless this is by design (and you think it's useful for the testing) it might be more obvious to readers if the queries don't lose the graph bindings via projection (i.e. consider changing the queries to use `SELECT *` and adjust results accordingly).

kasei (approved):

RobinTF:
@kasei I updated the tests according to your recommendation.

rubensworks (approved):

PR #375, opened 2026-09-23
SPARQL 1.1: `VALUES` inside `GRAPH` without the graph variable

RobinTF:
This is a PR that implements part of the tests proposed in #305. Adds a test checking that `VALUES` inside `GRAPH ?g` is joined with every named graph when `?g` is not mentioned inside. This conflicts with #376 and will be rebased once one of them is merged.


sparql/sparql11/bindings/graph-no-var.ttl
Tpt: hyper-nit: using the XML results format is a bit more readable

Tpt (approved):

sparql/sparql11/bindings/graph-no-var.ttl
RobinTF: It is done that way because `<data01.ttl>` clearly resolves against the filename of the turtle file.
Not sure if XML is standardized in the same way that makes this possible?
It's the same as https://github.com/w3c/rdf-tests/blob/369a90d1a60c021b746df2e411da0ff36258a758/sparql/sparql11/bindings/graph.ttl in the same directory when it comes to that.

Tpt: Good point! Let's keep it the test that way then

rubensworks (approved):

PR #376, opened 2026-09-23
SPARQL 1.1: `VALUES` inside `GRAPH` with `FROM NAMED`

RobinTF:
Adds tests checking that `GRAPH` around `VALUES` only yields graphs of the dataset defined by `FROM NAMED`, not excluded or unknown graphs. This conflicts with #375 and will be rebased once one of them is merged.


Tpt (commented):
I am not fully sure of the value of these tests. IF we have properly tested `FROM NAMED` + `GRAPH` behavior and `GRAPH` + `VALUES` behavior then it feels fairly fine to assume that the combination will be correct. WDYT?

kasei:
I think there's some value in having these tests simply because having `VALUES` enclosed by `GRAPH` without any other patterns tests a the particular case of producing values that are not coming from graph data. There's a bit of a slippery slope risk in testing every combination of features, but I think this one is valuable, and I've personally seen bugs in implementations due to the exact sort of thing being tested here.

As a minor comment, I would prefer to see expected results encoded using one of the standardized result formats (srj or srx) instead of the RDF-based ResultSet format.

Tpt:
@kasei Yes! There is already [a test](https://github.com/w3c/rdf-tests/blob/369a90d1a60c021b746df2e411da0ff36258a758/sparql/sparql11/bindings/graph.rq#L3) and #375 that is also covering `GRAPH` + `VALUES`. My point was on the `FROM NAMED` + `GRAPH` + `VALUES` combination.

> As a minor comment, I would prefer to see expected results encoded using one of the standardized result formats (srj or srx) instead of the RDF-based ResultSet format.

+1

kasei:
> @kasei Yes! There is already [a test](https://github.com/w3c/rdf-tests/blob/369a90d1a60c021b746df2e411da0ff36258a758/sparql/sparql11/bindings/graph.rq#L3) and #375 that is also covering `GRAPH` + `VALUES`. My point was on the `FROM NAMED` + `GRAPH` + `VALUES` combination.

Yeah, understood. I think my reasoning still stands (for me) in that testing the combination is valuable. I can see a clear difference between that query setup with and without the `FROM NAMED` clause, and have personal experience with bugs relating to this sort of combination. Like I said, there's a bit of a slippery slope thing here, and it's not critical, but I'd be happy to see this sort of test included.

Tpt:
> I can see a clear difference between that query setup with and without the FROM NAMED clause, and have personal experience with bugs relating to this sort of combination
>  Like I said, there's a bit of a slippery slope thing here, and it's not critical, but I'd be happy to see this sort of test included.

Make sense! +1 to all you said. Fine by me to add these tests.

kasei:
(I should clarify that my statements above were meant generally for the testing of the combination of features. I haven't closely reviewed the specifics of these tests, so am not approving the PR just yet.)

RobinTF:
@Tpt @kasei The reason why I proposed this test in the first place is that it actually exercises a unique code path in QLever. Happy to hear that @kasei also stumbled over this in the past, so it's not just us :D

Regarding the result format: How do you expect the graphs names to be resolved in srx/srj? The "nice" thing about turtle is that I can put in `<data01.ttl>` which then correctly resolves to the correct IRI, which on failure makes it obvious which graph failed. I could of course do a `?g = <data01.ttl>` in the SPARQL, so the result expectation would only need to contain "true"/"false", but that way we lose the actual graph.
It looks like newer SPARQL 1.2 tests use TriG to actually encode the graph names independently of the directory the files are in. Should I use that, even if it's a SPARQL 1.1 test?

rubensworks:
> It looks like newer SPARQL 1.2 tests use TriG to actually encode the graph names independently of the directory the files are in. Should I use that, even if it's a SPARQL 1.1 test?

I don't remember us changing anything to this specifically for the SPARQL 1.2 tests, so it should be safe to also do this in the 1.1 test suite.

PR #377, opened 2026-09-23
SPARQL 1.1: add `LIMIT`/`OFFSET` subquery inside `GRAPH` tests

RobinTF:
This is a PR that implements part of the tests proposed in #305. Adds tests checking that a subquery with `ORDER BY`/`OFFSET`/`LIMIT` inside `GRAPH ?g` is evaluated separately for each named graph.


Tpt (approved):
thanks! 

Micro nit: I would maybe not set `dawgt:approval dawgt:Proposed ;` because these tests are not targeting any existing W3C WG new specification. But it's my personal opinion, not sure if the other group chairs agree

RobinTF:
@Tpt I have no opinion on that. I let the AI write the manifest files for me and it looks like it decided to mimic the existing tests like that.

kasei:
The `dawgt:Proposed` modeling was the style used in the 1.0 and 1.1 working groups. I think we're probably moving to a state where the approval comes as part of the merging in git via PR, so I tend to agree that it doesn't make sense for new tests.

kasei (approved):

rubensworks (approved):
Looks good, thanks!

PR #378, opened 2026-09-23
SPARQL 1.1: `STRLANG()` with language tags differing in case

RobinTF:
This is a PR that implements part of the tests proposed in #305. Adds a test checking that `STRLANG()` results with language tags differing only in case are equal.


kasei (approved):

RobinTF:
After the discussion on #379 I believe this PR could suffer from the same problem. I could fold these two expressions into #379 and close this PR. Would you prefer that, or keeping it as is?

rubensworks (approved):

PR #379, opened 2026-09-23
SPARQL 1.0: `LANG()` with language tags differing in case

RobinTF:
This is a PR that implements part of the tests proposed in #305. Adds a test checking that `LANG()` returns equal values for language tags differing only in case.


afs:
The way language tags are handled is different between RDF and hence SPARQL versions.

RDF 1.1 --
> Lexical representations of language tags MAY be converted to lower case. The value space of language tags is always in lower case.

Note that `LANG = LANG` is a `xsd:string` comparison.

RDF 1.2 --
> [ASCII case-insensitive matching](https://www.w3.org/TR/i18n-glossary/#dfn-ascii-case-insensitive)

I think it is better to put language tag tests in SPARQL 1.2 (if not there), and leave 1.0, 1.1 alone.

A language tag comparison would need a graph pattern (although `"abc"@en = "abc"@EN` is likely to work).

So a pure way to check is to be a term match in the graph.

Data:
```
:x :p "abc"@en .
:x :q "abc"@EN .
```
and pattern
```
WHERE { 
   ?x :p ?v .
   ?x :q ?v .
}
```

A related issue is that RFC 5646 describes a different standard representation of language tags. The "language" subtag is lowercase in both.

RobinTF:
@afs Thank you for your feedback. So your proposal is simply to move these tests to the sparql 1.2 directory and keep it as-is?

afs:
I haven't looked at the other `LANG` related tests in detail yet.

For this one, the portable way in SPARQL 1.2 is the graph pattern match style.

FWIW: Some triple stores canonicalize terms to RFC4656 format, e.g. `en-US`, some to lowercase, `en-us`. Neither is wrong.

The graph pattern way to test means it is "case-insensitive matching". `LANG` is an xsd:string so the comparison is then case sensitive.

sparql/sparql10/expr-builtin/lang-case-insensitive-lang.srx
```diff
@@ -0,0 +1,98 @@
...
+      <variable
+       name="v1" />
+      <variable
+       name="x2" />
+      <variable
+       name="v2" />
+   </head>
+   <results
+   
+   >
```
kasei: nit: strange use of whitespace here

sparql/sparql10/expr-builtin/lang-case-sensitivity-lang.rq
```diff
@@ -0,0 +1,10 @@
+# Test: LANG('xyz'@en) = LANG('xyz'@EN)
+
+PREFIX     :    <http://example/>
+
+SELECT *
+{
+    ?x1 :p ?v1 .
+    ?x2 :p ?v2 .
+    FILTER ( LANG(?v1) = LANG(?v2) )
```
kasei: I don't think this is the right way to do this sort of testing. Depending on the SPARQL system, normalization might have already happened by the time the dataset is constructed (before the query even starts). If that happens, this `LANG` comparison becomes a no-op. To test the behavior of `LANG` at runtime, a better approach might be to construct the strings at runtime using `STRLANG`. Some systems might perform canonicalization in the `STRLANG` constructor as well, but at least you'd have some confidence that *query runtime behavior* was treating the case differences as equivalent.

RobinTF:
@afs @kasei Per your suggestion I changed the tests to target SPARQL 1.2. Let me know what you think of the new proposal.

rubensworks (approved):
Comunica currently fails on these tests, but the tests themself look correct to me.

kasei (approved):

Tpt (approved):

